Issue #158, opened 2026-04-04
Reassess LiteLLM version cap after dismissed Dependabot alert #25


duriantaco:
Skylos currently caps `litellm` at `<1.82.8` for prior security / supply-chain reasons.

Dependabot alert #25 (GHSA-jjhc-v7c2-5hh6 / CVE-2026-35030) was dismissed by maintainer as `not_used` because we do not use the affected LiteLLM proxy / JWT auth path. However, opening an issue to reassess the LiteLLM version cap and decide in the future on a safe upgrade strategy for optional LLM support.

## Context:
  - alert: #25
  - package: `litellm`
  - current cap: `>=1.81.3,<1.82.8`
  - current usage in Skylos appears to be LiteLLM as a client library, not the vulnerable JWT/OIDC auth cache flow

## Follow-up tasks:
  - review LiteLLM releases after 1.82.8, including 1.83.0+
  - confirm whether the previous cap can be relaxed safely
  - test Skylos LLM flows against a newer LiteLLM version
  - decide whether to raise, replace, or keep the cap

duriantaco:
Resolved by #198. The LiteLLM cap was relaxed to `<1.84.0`, the lockfile was refreshed to LiteLLM 1.83.7, and the targeted validation passed against that updated dependency set.

PR #198, opened 2026-04-13
chore(llm): relax litellm version cap

mikepatraw:
## What does this PR do?

Relaxes Skylos's LiteLLM upper bound from `<1.82.8` to `<1.84.0`.

## Why?

Issue #158 asks whether the current cap is still necessary after the dismissed Dependabot alert.

I reviewed the LiteLLM usage in Skylos and did not find any repo-specific incompatibility that justified keeping the stricter cap. The previous bound looked precautionary, but I could not find validation showing that newer 1.83.x releases break Skylos's adapter or LLM verification flow.

## How to test

- `UV_PYTHON=python3.13 uv sync --extra llm --extra test`
- `UV_PYTHON=python3.13 uv run pytest test/test_litellm_adapter.py test/test_cli_llm_provider.py test/test_llm_schemas.py test/test_verify_orchestrator.py -q`

Local result: `148 passed`

## Precision Impact

None. This only updates the optional LiteLLM dependency range.

## Checklist

- [x] Tests pass for the dependency path touched here
- [x] No analysis logic changed
- [x] Corpus update not applicable
- [x] Corpus guard not applicable
- [ ] CHANGELOG updated


duriantaco (commented):
Thanks for the PR.. but the current test evidence doesnt show that we actually exercised a newer LiteLLM release.

Lockfile resolves `litellm` to `1.81.15`, which actually still satisfies both the old range (`>=1.81.3,<1.82.8`) and the new range (`>=1.81.3,<1.84.0`). So the `uv sync --extra llm --extra test` command can still end up testing 1.81.15 rather than any 1.83.x version.

Can you either 1. update the lockfile to a real 1.83.x LiteLLM version and rerun the targeted tests, or 2. explicitly install/test against 1.83.x in the validation steps?

Thanks!

RedParrotBerkeley:
Updated this to address your review comment.

The lockfile now resolves LiteLLM 1.83.7, and I reran the targeted validation against that locked state.

Tested with:
- `UV_PYTHON=python3.13 uv sync --extra llm --extra test`
- `UV_PYTHON=python3.13 uv run pytest test/test_litellm_adapter.py test/test_cli_llm_provider.py test/test_llm_schemas.py test/test_verify_orchestrator.py -q`

Local result: `148 passed`

duriantaco:
Thanks for the update! Pulled the PR locally, synced the locked env with Python 3.13, and reran the targeted suite against LiteLLM 1.83.7. 148 passed so all's good. 

One non-blocking follow up though: CI still doesnt install/test the `llm` extra in test workflow... So may be worth adding that coverage separately.

PR #201, opened 2026-04-14
ci(test): install llm extra in test workflow

mikepatraw:
## What does this PR do?

Updates the main test workflow to install the `llm` extra before running the test suite.

## Why?

Follow-up to #198.

That PR validated Skylos against a newer LiteLLM release, but the existing `Tests + Coverage` workflow still installs `-e .` without `.[llm]`. That means CI can pass without exercising the optional LiteLLM dependency path.

This change makes the workflow cover the same optional dependency surface that the LiteLLM-related tests and adapter path depend on.

## How to test

- Review `.github/workflows/tests.yaml`
- Confirm the workflow now installs:
  - `uv pip install -e ".[llm]"`
- Let GitHub Actions run the updated workflow on this PR

## Precision Impact

None. This only updates CI dependency installation.

## Checklist

- [x] No analysis logic changed
- [x] No corpus update needed
- [x] No corpus guard needed
- [x] CHANGELOG not needed

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Create venv + install deps
         run: |
           uv venv
-          uv pip install -e .
+          uv pip install -e ".[llm]"
           uv pip install pytest pytest-cov
 
       - name: Run tests with coverage
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -9,6 +9,12 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        install_target:
+          - "."
+          - ".[llm]"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


duriantaco:
Thanks for the follow-up on #198! I think we might need to change this a little though. Switching the existing test workflow to .[llm] will mean that we lose our main pytest coverage for the default install path. We plan to keep LiteLLM optional (at least for the foreseeable future). Would you be open to keeping the current test job as is and adding a second job or matrix entry for uv pip install -e ".[llm]" instead? I think that would give us coverage for the LiteLLM path without losing coverage for the base install.

duriantaco (dismissed):

duriantaco:
CI is green already. The PR is blocked because branch protection still expects a check called \test, but this workflow now emits `test (.)` and `test (.[llm])`. Can you add a final aggregator job called `test` that depends on both matrix jobs?  If you want, I can patch it on your branch. Do lemme know

duriantaco:
Anyways, patched the workflow on your branch to keep both matrix installs. CI all green now. Thanks a lot for your help! Greatly appreciate it.

RedParrotBerkeley:
Thanks, really appreciate you patching that and keeping both install paths covered.

PR #296, opened 2026-05-05
fix(architecture): preserve submodule coupling targets

### LINKED ISSUE #293: Architecture metrics: are inter-package imports intentionally aggregated to root packages? (Q802/Q803 on submodules)
Hi! First, thanks for skylos — using it on a multi-CLI Python repo and the gate workflow is great.

I have a question about the **architecture metrics** (`SKY-Q802` / `SKY-Q803`) and how `Ca` (afferent coupling) is computed for **submodules** of a package. I want to make sure I understand whether the current behaviour is intentional or a latent bug, before I either work around it or open something more concrete.

## TL;DR question

When module `A.cli` (a submodule of package `A`) is imported by another module `B.cli` via any of:

- `from A.cli import main`
- `import A.cli`
- `import A.cli as foo`
- `from A import main` (re-exported via `A/__init__.py`)

…**none of these increment `A.cli.ca`**. They only increment `A.ca` (the package root). Is this intentional? If so, why do `module_metrics` then report **per-submodule** entries (e.g. `A.cli`) with `ca=0` that mechanically end up in `zone_of_uselessness`?

## Setup that triggered the question

```
src/
  package_a/
    __init__.py          # `from package_a.cli import main`
    cli.py               # imports `sync_common`
  package_b/
    __init__.py
    cli.py               # imports `sync_common` AND `from package_a.cli import main as run_a`
  sync_common.py         # shared helpers
```

`package_b.cli` explicitly imports `from package_a.cli import main`. I expected `package_a.cli.ca = 1`. Skylos reports `package_a.cli.ca = 0` and `package_a.ca = 1`.

## Empirical reproducer

I tried 4 import styles in `package_b/cli.py`, all produced identical metrics:

| Import in `package_b/cli.py`             | `package_a.ca` | `package_a.cli.ca` |
|------------------------------------------|----------------|--------------------|
| `from package_a.cli import main as run_a`| 1              | **0**              |
| `import package_a.cli as _mod`           | 1              | **0**              |
| `import package_a.cli`                   | 1              | **0**              |
| `from package_a import main as run_a`    | 1              | **0**              |

Consequence: `package_a.cli` and `package_b.cli` end up with `ca=0, ce>=1, I=1.0, A≈0.05` → `zone_of_uselessness` → `SKY-Q803` finding.

## Code paths I think are responsible

In `skylos/circular_deps.py`, all import targets are reduced to the root package via `.split(\".\")[0]`:

\`\`\`python
# circular_deps.py:54 (visit_Import)
module = alias.name.split(\".\")[0]
...
to_module=alias.name.split(\".\")[0],

# circular_deps.py:68 (visit_ImportFrom)
module = node.module.split(\".\")[0]
...
to_module=module,

# circular_deps.py:108 (build_graph_from_raw_imports)
root = import_module.split(\".\")[0]
if root in self.known_modules:
    dep = ModuleDependency(..., to_module=root, ...)
\`\`\`

This `dependency_graph` is then consumed by `architecture.analyze_architecture()`:

\`\`\`python
# architecture.py:182
for module, deps in dependency_graph.items():
    for dep in deps:
        if dep in all_modules and dep != module:
            efferent[module].add(dep)
            afferent[dep].add(module)
\`\`\`

Since `dep` is always the root package, `afferent[\"package_a.cli\"]` is never populated even though Python-wise the import semantically targets `package_a.cli`.

But then findings are **emitted at the submodule level** with the submodule simple name:

\`\`\`python
# architecture.py:401, 439, 461
\"name\": name,
\"simple_name\": name.split(\".\")[-1],
\`\`\`

So a user sees `plugins_sync.cli` flagged as `zone_of_uselessness`, drills into the imports, and is confused because semantically modules **do** import that submodule.

## What I'd like to clarify

1. **Is the root-package aggregation intentional?** I couldn't find it documented in [docs.skylos.dev/rules-reference](https://docs.skylos.dev/rules-reference) or anywhere else.
2. **If intentional**, would it make sense to also report metrics at the package level only (avoid showing `package.cli` separately when its `Ca/Ce` aren't really computed for it)?
3. **If unintentional**, would you accept a PR that:
   - tracks the full dotted path in `ModuleDependency.to_module` (e.g. `package_a.cli` instead of `package_a`),
   - keeps a `to_root` derived field for cycle detection if needed?

I'm happy to send a PR if you'd like — just want to confirm the design intent first.

Skylos version: `4.9.0` (installed via `uv sync`).
Python: `3.14.3`.

Thanks!

duriantaco:
## Summary

Fixes architecture metrics for explicit dotted imports to submodules. Keep the existing root collapsed graph for circular dependency detection, and add architecture graph for `Ca/Ce` metrics.

## Changes

  - Kept existing circular dependency behavior with root collapsed edges.
  - Add architecture dependencies for known dotted modules.
  - Use precise graph when computing architecture metrics

## Verification

```
pytest test/test_circular_deps.py test/test_architecture.py
```

Fixes #293.

### DIFF
--- a/skylos/analyzer.py
+++ b/skylos/analyzer.py
@@ -1563,7 +1563,9 @@ class Skylos:
                     try:
                         from skylos.architecture import get_architecture_findings
 
-                        dep_graph = dict(circular_rule._analyzer.dependencies)
+                        dep_graph = dict(
+                            circular_rule._analyzer.architecture_dependencies
+                        )
                         mod_files = dict(circular_rule._analyzer.modules)
 
                         mod_trees = {}
--- a/skylos/circular_deps.py
+++ b/skylos/circular_deps.py
@@ -10,6 +10,47 @@ except ImportError:
     _fast_find_cycles = None
 
 
+def _module_root(module_name: str) -> str:
+    return module_name.split(".")[0] if module_name else ""
+
+
+def _known_module_names(module_name: str) -> Set[str]:
+    if not module_name:
+        return set()
+    return {module_name, _module_root(module_name)}
+
+
+def _resolve_known_module(module_name: str, known_modules: Set[str]) -> str | None:
+    parts = module_name.split(".")
+    for end in range(len(parts), 0, -1):
+        candidate = ".".join(parts[:end])
+        if candidate in known_modules:
+            return candidate
+    return None
+
+
+def _resolve_from_import_targets(
+    import_module: str, imported_names: List[str], known_modules: Set[str]
+) -> Dict[str, List[str]]:
+    base_target = _resolve_known_module(import_module, known_modules)
+    targets: Dict[str, List[str]] = defaultdict(list)
+
+    if not imported_names:
+        if base_target:
+            targets[base_target] = []
+        return dict(targets)
+
+    for imported_name in imported_names:
+        member_target = _resolve_known_module(
+            f"{import_module}.{imported_name}", known_modules
+        )
+        target = member_target or base_target
+        if target:
+            targets[target].append(imported_name)
+
+    return dict(targets)
+
+
 @dataclass
 class ModuleDependency:
     from_module: str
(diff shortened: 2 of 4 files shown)


Flamintus:
@duriantaco Wow, how reactive! I'm going to test version 4.11 right now. THX !!

PR #325, opened 2026-05-09
fix(gate): make file-level IAD architecture findings advisory

### LINKED ISSUE #322: Q802/Q803 under `--strict`: scope mismatch with Martin's original release-unit granularity
## Context

First, sincere thanks for the work done on Q802/Q803 over the last couple of weeks. The contextual exemption layer that landed in 4.11→4.12 (PR #296 for submodule resolution, PR #301 for underscore + entrypoint, PR #307 for `__init__.py` re-export, the 4.12.1 fix for #311) makes the rules considerably more usable on idiomatic Python layouts. We've watched the iterations closely and the direction is good.

This issue is **not** a false-positive report. The latest exemption logic (4.12.1) is doing what it's supposed to. Instead, we want to document — for the maintainer and for future users hitting the same wall — **why we've decided to keep Q802/Q803 disabled in our gate** despite the recent fixes, and why we believe these two rules should not be enforced atomically at the file level under `--strict`.

## TL;DR

Robert C. Martin's `I/A/D` metric (*Agile Software Development*, 2002) is defined for **release-units** — JARs, Maven artifacts, .NET assemblies, npm packages versioned independently — under the explicit principle *"the granule of reuse is the granule of release"*. Skylos applies the same formula at the granularity of an **individual `.py` file inside a single release-unit**. That is a scope shift Martin's framework was not built for, and it produces structural false positives that no exemption ruleset can fully neutralize without effectively disabling the rule.

For our project (`go-my-ai-agent-sync`, a single `uv_build` package distributed as one PyPI artifact), this means:

- A correctly designed internal helper (`_helpers.py`: 4 pure functions, 6 internal consumers, zero external dependency) sits permanently in the Zone of Pain by mathematical construction (`Ca > 0`, `Ce = 0`, `A ≈ 0`).
- The `private_helper_ca_limit = 3` exemption (`architecture.py:515`) handles small helpers, but on a 6-module CLI the central helper crosses the threshold structurally — `get_logger` alone has `Ca = 6`.
- The available remediations from the finding message — *"introduce stable abstractions"* / *"reduce fan-in"* — push toward either an **anti-pattern** (`Protocol` with one implementer = Header Interface, [Drotbohm 2024](http://odrotbohm.github.io/2024/09/the-instability-abstractness-relationsship-an-alternative-view/)) or a **disproportionate restructure** (split a 4-function helper into 4 sub-modules whose individual `Ca` would still cross the threshold because the same consumers use all of them).

## What the metric is for, in Martin's own framing

From *Agile Software Development*, the principle accompanying `I/A/D` is the Reuse/Release Equivalence Principle:

> The granule of reuse is the granule of release. Only components that are released through a tracking system can be effectively reused.

Martin's canonical example for the Zone of Pain is **a database schema** — a heavyweight, slowly-evolving foundation shared by many independently-released components. The metric is meant to flag a release-unit whose stability and concreteness combination makes coordinated migrations across consumers expensive (think: a JAR depended upon by dozens of other JARs in a multi-module Maven build).

Inside a single release-unit, that coordination cost does not exist. A change to `_helpers.py` is released atomically with its 6 importers, in the same commit, in the same `0.2.x` version. There is no "rippling change cost across consumers" the metric is trying to surface, because there are no independent consumers.

## Why no amount of contextual exemptions can fully fix this

The exemption logic in `_filter_contextual_findings` (`architecture.py:569`) is essentially a list of **escape hatches for known idioms** — `_` prefix, entrypoint, `__init__.py` re-export, test module, package-boundary re-export. Each one neutralizes a specific shape of structural false positive. But the underlying issue is mathematical:

- A well-designed helper inside a release-unit has `Ce = 0` (no internal upstream) and `A ≈ 0` (concrete code, by definition of "helper"). Therefore `I = 0` and the module is in the Zone of Pain *by construction*. This is true for `os.path`, `pathlib`, `collections.abc` at the stdlib scale and for any internal `_helpers.py` at the project scale.
- The only way to make such a module *pass* the rule under `--strict` is to either (a) raise `A` artificially via single-implementer `Protocol`s — which Drotbohm and Fowler both classify as anti-patterns — or (b) move the helper out of the release-unit and version it separately, which is wildly disproportionate for a 4-function file.
- The exemption list is therefore unbounded: every new idiomatic Python layout (Typer dispatcher, FastAPI router, Pydantic settings module, `__main__.py`, `_internal/_*` private subpackages) eventually surfaces a new "false positive" that needs its own escape hatch.

This is not a critique of the exemption work — it has been excellent — but a structural observation: as long as `--strict` treats `Q802/Q803` as binary blocking findings, the maintenance burden of keeping the exemption list complete is open-ended, and projects keep hitting the wall after each new fix.

## Modern critique of the metric itself

The most pointed contemporary critique is Oliver Drotbohm (Spring engineering team), [*The Instability-Abstractness-Relationship — An Alternative View*](http://odrotbohm.github.io/2024/09/the-instability-abstractness-relationsship-an-alternative-view/), 2024:

1. **Abstractness ≠ Abstraction.** Extracting an interface from a class raises the metric numerically without improving semantic abstraction or change resistance.
2. **Misaligned with real change patterns.** Martin assumed extension-via-interface as the dominant change driver. Real business-application changes are typically *introducing new abstractions* or *adding methods to existing types*, both of which lower the metric without harming design.
3. **Incentivizes meaningless refactoring.** Developers (and LLM agents — see issue #320) add interfaces purely to satisfy the score.
4. **Questionable design assumption.** The "extensibility through abstraction" model fits collection libraries and framework extension points, not typical business code.

His conclusion: the I/A relationship *"does not support change patterns of modern applications"* and should be reconsidered as an architectural decision driver.

## Decision on our side

We are keeping `SKY-Q802` and `SKY-Q803` **disabled** in the local pre-commit gate and in CI for `go-my-ai-agent-sync`. Concretely:

- `--strict` is preserved for every other Skylos rule family (security, secrets, dead code, danger, SCA) and for every other quality tool in the gate (ruff, ty, basedpyright, pylint, gitleaks, pip-audit, pytest).
- Q802 and Q803 are excluded specifically because their underlying metric is calibrated for a granularity that does not match what we are scanning.

This is a **calibration** decision, not a bypass: we still want Skylos in the loop for the gate-shaped findings; we just stop treating an architectural-shape indicator as a binary blocker.

## Suggestions, in priority order

These are not asks — they are options the maintainer might consider if/when revisiting Q802/Q803:

1. **Reclassify Q802/Q803 as advisory rather than gate-blocking by default.** Document them as architectural indicators; require an explicit opt-in (e.g. `--strict-architecture` or a config key) to make them blocking. This aligns with Martin's own framing — these are signals to discuss in design review, not test failures.
2. **Document the intended granularity.** A short paragraph in the rules reference clarifying that `I/A/D` was originally defined for release-units, and that Skylos applies it at file granularity as a heuristic, would set expectations correctly and reduce the "why is my well-designed helper red?" friction.
3. **Consider a `--per-package` mode.** Aggregate metrics at the package root rather than the module. For a single-package release-unit this would emit at most one Q802/Q803 finding per top-level package, which is closer to the original framing.
4. **Cite Drotbohm's critique** in the rule documentation as a known limitation. Acknowledging that the metric is contested in the modern Java/Spring ecosystem helps users calibrate how much weight to give it.

## Reference material

- The 7-fixture empirical mapping at [`Flamintus/test_skylos_Q802-3`](https://github.com/Flamintus/test_skylos_Q802-3) (referenced in #320) — fixture 02 (`02-private-helper`) is the minimal "structural Pain by construction" case.
- Martin, *Agile Software Development: Principles, Patterns, and Practices*, 2002, ch. 20 (Reuse/Release Equivalence Principle).
- [Drotbohm, *The Instability-Abstractness-Relationship — An Alternative View*, 2024](http://odrotbohm.github.io/2024/09/the-instability-abstractness-relationsship-an-alternative-view/).
- [Software package metrics — Wikipedia](https://en.wikipedia.org/wiki/Software_package_metrics).

## What we hope this issue accomplishes

- Document the trade-off so the next user hitting Q803 on a legitimate helper finds a public reference rather than reaching for `noqa` or rewriting their code into Header Interfaces.
- Open the conversation on whether `Q802/Q803` belong under the same `--strict` umbrella as binary correctness rules, or whether they are a different class of signal.
- Acknowledge — explicitly — that the recent fixes are good work; the disable decision is about the metric's scope, not the implementation quality.

Thanks again for the project and the responsiveness.

duriantaco:
Fixes #322

## Problem

`SKY-Q802` and `SKY-Q803` use Martin’s instability/abstractness/distance metric which was meant for release units, Skylos applies it at Python file level, which can be too strong as a strict gate failure.

## Fix

Make file-level Q802/Q803 findings advisory by default. Gatekeeper now excludes advisory Q802/Q803 findings from strict-gate and quality threshold failure counts.

Projects that want the old blocking behavior can opt in:

  ```toml
  [tool.skylos.architecture]
  enforce_iad = true
```

## Verification

```
 pytest -q test/test_architecture.py test/test_gatekeeper.py test/test_gate_kwargs.py test/test_analyzer.py test/test_cli.py test/test_cli_precommit.py
```

### DIFF
--- a/skylos/analyzer.py
+++ b/skylos/analyzer.py
@@ -144,6 +144,15 @@ def _entrypoint_module_name(qname: str) -> str | None:
     return module_name or None
 
 
+def _architecture_iad_strict(architecture_cfg) -> bool:
+    if not isinstance(architecture_cfg, dict):
+        return False
+    for key in ("enforce_iad", "strict_iad"):
+        if key in architecture_cfg:
+            return bool(architecture_cfg.get(key))
+    return False
+
+
 def _expand_reexported_entrypoint_modules(
     entrypoint_qnames: set[str],
     entrypoint_modules: set[str],
@@ -1683,6 +1692,9 @@ class Skylos:
                             entrypoint_modules=entrypoint_modules,
                             package_boundary_modules=package_boundary_modules,
                             layer_policy=project_cfg.get("architecture"),
+                            iad_findings_advisory=not _architecture_iad_strict(
+                                project_cfg.get("architecture")
+                            ),
                         )
                         ignored_rules = set(project_cfg.get("ignore", []))
                         if arch_findings:
--- a/skylos/architecture.py
+++ b/skylos/architecture.py
@@ -194,6 +194,7 @@ def analyze_architecture(
     module_trees: dict[str, ast.AST] | None = None,
     module_abstractness: dict[str, dict[str, Any]] | None = None,
     module_loc: dict[str, int] | None = None,
+    iad_findings_advisory: bool = True,
 ) -> ArchitectureResult:
     result = ArchitectureResult()
 
@@ -407,12 +408,50 @@ def analyze_architecture(
             "zone_distribution": {},
         }
 
-    _generate_findings(result)
+    _generate_findings(result, iad_findings_advisory=iad_findings_advisory)
 
     return result
 
 
-def _generate_findings(result: ArchitectureResult):
+def _iad_scope_fields(advisory: bool) -> dict[str, Any]:
+    fields = {
+        "advisory": advisory,
+        "scope": "file",
+        "metric_granularity": "file-level heuristic",
+        "metric_origin": "Martin I/A/D release-unit metric",
+    }
+    if advisory:
+        fields["advisory_reason"] = (
+            "Martin's instability/abstractness/distance metric was defined for "
+            "release units. Skylos reports it at Python file granularity as an "
+            "architecture signal; it is not gate-blocking unless I/A/D enforcement "
+            "is enabled."
+        )
+    else:
+        fields["enforcement_reason"] = (
+            "I/A/D enforcement is enabled, so this file-level architecture signal "
+            "can block strict gates."
+        )
+    return fields
+
+
+def _iad_gate_note(advisory: bool) -> str:
+    if advisory:
+        return (
+            "Advisory: this file-level I/A/D signal does not block gates unless "
+            "I/A/D enforcement is enabled."
+        )
+    return (
+        "I/A/D enforcement is enabled, so this file-level architecture signal can "
+        "block strict gates."
+    )
+
+
+def _generate_findings(
+    result: ArchitectureResult,
+    *,
+    iad_findings_advisory: bool = True,
+):
     for name, m in result.modules.items():
         # SKY-Q802: High distance from main sequence
         if m.distance > 0.5 and m.total_coupling > 0:
@@ -436,11 +475,13 @@ def _generate_findings(result: ArchitectureResult):
                     "message": (
                         f"Module '{name}' is far from the Main Sequence "
                         f"(D={m.distance:.2f}, I={m.instability:.2f}, A={m.abstractness:.2f}). "
-                        f"Zone: {m.zone.replace('_', ' ')}."
+                        f"Zone: {m.zone.replace('_', ' ')}. "
+                        f"{_iad_gate_note(iad_findings_advisory)}"
                     ),
                     "file": m.file_path,
                     "basename": Path(m.file_path).name if m.file_path else name,
                     "line": 1,
+                    **_iad_scope_fields(iad_findings_advisory),
                 }
             )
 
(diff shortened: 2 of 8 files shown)


Flamintus:
I sincerely believe this is the right decision. It seems like the right thing to do.

duriantaco:
Thanks for all the awesome help 👍  Do feel free to submit a PR next time if you're confident! Also do feel free to raise any issues or feature requests any time. Will be more than happy to help in any way possible

Flamintus:
@duriantaco Closing thoughts — one last thanks 🙏

With Q802/Q803 advisory by default in 4.13, the practical impact is exactly what we hoped for: **no more false positives bleeding into the gate**, and just as importantly, LLM coding agents no longer reach for cosmetic restructures (or fabricated abstractions / re-exports) just to silence the rule. The signal stays where it belongs — informational, in the report — and the gate stays honest.

A direct consequence we already see: it becomes possible again to **organize code cleanly by responsibility**, even when a private `_xxx.py` is only consumed by a single `core.py`. Splitting for readability and maintainability is no longer punished by a `Ca/Ce` shape the metric was never meant to evaluate at file granularity.

Small housekeeping note: the upstream repo I built for the fixtures (`test_skylos_Q802-3` — the `01`–`08` layouts and the audit reports) will stay public for **about one more week** before I archive it, in case any of it is still useful on your end.

For my part, I'm moving from Python to JavaScript/TypeScript stacks for the next stretch of work. Since Skylos covers JS/TS too, I'll likely keep running it on the new codebases — so I won't hesitate to come back if I spot anything worth raising from that side as well. 🙂

Thanks again for the responsiveness and the willingness to engage on design discussions, not just bug reports. It made the iteration genuinely worthwhile.

PR #333, opened 2026-05-10
fix: restrict agent service CORS to trusted origins instead of wildcard

andesyteoss:
## Summary

The agent service HTTP handler in `skylos/agent_service.py` returns `Access-Control-Allow-Origin: *` on every response, which makes its endpoints reachable cross-origin from any website the user visits while the service is running. This PR replaces the wildcard with an origin allow-list (defaulting to `http://127.0.0.1:<port>` and `http://localhost:<port>`, configurable via a new `allowed_origins` parameter).

CWE-352 (CSRF) / CWE-942 (Permissive CORS).

## Affected code

`skylos/agent_service.py::AgentServiceHandler._send_cors_headers` (pre-fix):

```python
def _send_cors_headers(self) -> None:
    self.send_header("Access-Control-Allow-Origin", "*")
    self.send_header("Access-Control-Allow-Headers", "Content-Type, X-Skylos-Agent-Token")
    self.send_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
```

The handler exposes state-changing endpoints (`dismiss`, `snooze`, `learn`, `clear`, `refresh`) that mutate the on-disk agent state.

## Why this is exploitable

The service is intended to run locally, but `Access-Control-Allow-Origin: *` opts every browser-reachable client into being able to read responses cross-origin. Two scenarios:

1. **Service started without a token** (programmatic/`create_agent_service` callers that don't pass `token=...`). `_is_authorized` returns `True` when `token is None`, so any page the user visits can issue `fetch()` calls to `http://127.0.0.1:<port>/...` — preflight succeeds because of the wildcard, and the page can read and mutate agent state freely.
2. **Service started with a token** (the CLI default). The `X-Skylos-Agent-Token` custom header is required, which a cross-origin page cannot guess (the CLI generates a 192-bit `secrets.token_urlsafe(24)`). Practical CSRF is blocked here, but the wildcard still unnecessarily widens the attack surface and would expose any future endpoint that doesn't enforce the header.

`Access-Control-Allow-Credentials` was never set, so cookie-based attacks aren't in scope; the risk is specifically header-auth and tokenless modes.

## Fix

- `_send_cors_headers` now reads the request `Origin` and only echoes it back (and sends the matching `Allow-Headers`/`Allow-Methods`) when it appears in an explicit allow-list. Same-origin and non-browser callers are unaffected because they don't rely on these headers.
- `create_agent_service` accepts a new `allowed_origins: list[str] | None` parameter. When omitted, defaults are derived from the bind address: localhost variants for `0.0.0.0`/`127.0.0.1`/`::1`/`localhost`, otherwise `http://<host>:<port>`. When `port=0` is used, the server is bound first so the real port is reflected in the defaults.
- No behavior change for legitimate same-origin clients (the VS Code extension and any localhost UI already speak to `http://127.0.0.1:<port>`); operators who proxy through another origin can pass `allowed_origins=[...]` explicitly.

## Tests

Added `test/test_cwe352_agent_service_cors.py` with 5 cases covering:

- No `Access-Control-Allow-Origin` header is emitted for a disallowed origin (`http://evil.example`).
- Allowed origins are echoed back exactly (no wildcard).
- `OPTIONS` preflight from a disallowed origin doesn't grant CORS access.
- `OPTIONS` preflight from an allowed origin returns the expected `Allow-Headers` / `Allow-Methods`.
- Custom `allowed_origins` values supplied by callers are honored.

```
$ python3 -m pytest test/test_cwe352_agent_service_cors.py -q
.....                                                                    [100%]
5 passed in 2.62s
```

## Adversarial review

Before submitting, I tried to disprove this. The strongest counter-argument is that the CLI always generates a random `X-Skylos-Agent-Token`, and a cross-origin page can't read that token, so CSRF is already blocked in the default deployment. That's true and I want to be upfront about it — the practical severity for CLI users is low. However, `_is_authorized` permits unauthenticated access when `token is None` (callers using `create_agent_service` directly may run in this mode), and the wildcard CORS in that mode does allow any visited webpage to read and mutate agent state on `127.0.0.1`. The fix also closes the failure mode where a future endpoint forgets to enforce the token header. This is defense-in-depth that aligns the service with standard browser security practice for local-bound HTTP APIs.

## Notes

I noticed `SECURITY.md` prefers private disclosure. I went with a public PR here because (a) the default CLI deployment is mostly mitigated by the random token, so this is principally a hardening change, and (b) the patch itself is the fix and there's no separate exploit to coordinate. Happy to move discussion to `aaronoh2015@gmail.com` if you'd prefer.

### DIFF
--- a/skylos/agent_service.py
+++ b/skylos/agent_service.py
@@ -22,6 +22,17 @@ def _error_payload(message: str) -> dict[str, str]:
     return {"error": message}
 
 
+def _default_allowed_origins(host: str, port: int) -> list[str]:
+    """Build the default set of allowed CORS origins for the listening address."""
+    origins: list[str] = []
+    if host in ("0.0.0.0", "127.0.0.1", "::1", "localhost", ""):
+        origins.append(f"http://127.0.0.1:{port}")
+        origins.append(f"http://localhost:{port}")
+    else:
+        origins.append(f"http://{host}:{port}")
+    return origins
+
+
 class AgentServiceController:
     def __init__(
         self,
@@ -240,6 +251,7 @@ class AgentServiceHandler(BaseHTTPRequestHandler):
     controller: AgentServiceController
     token: str | None = None
     default_limit: int = 10
+    allowed_origins: set[str] = frozenset()
 
     def do_OPTIONS(self) -> None:
         self.send_response(HTTPStatus.NO_CONTENT)
@@ -321,11 +333,14 @@ class AgentServiceHandler(BaseHTTPRequestHandler):
         self.wfile.write(data)
 
     def _send_cors_headers(self) -> None:
-        self.send_header("Access-Control-Allow-Origin", "*")
-        self.send_header(
-            "Access-Control-Allow-Headers", "Content-Type, X-Skylos-Agent-Token"
-        )
-        self.send_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
+        origin = self.headers.get("Origin", "")
+        if origin and origin in self.allowed_origins:
+            self.send_header("Access-Control-Allow-Origin", origin)
+            self.send_header(
+                "Access-Control-Allow-Headers",
+                "Content-Type, X-Skylos-Agent-Token",
+            )
+            self.send_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
 
 
 class SafeAgentServiceHandler(AgentServiceHandler):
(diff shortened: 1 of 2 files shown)


duriantaco (commented):
Check `security.md`

duriantaco:
Thanks @sebastiondev for reporting this. We reimplemented the remediation in tree with stricter origin rejection, additional API hardening, and regression tests in #335. Since `SECURITY.md` asks that security issues be reported privately, we will closing this public PR rather than merging it directly. Thanks once again for your report.

PR #691, opened 2026-08-04
fix(ai-defect): recognize module dunder attributes in phantom refs

### LINKED ISSUE #684: FP: SKY-L012 on import-provided attributes
**What was flagged?**
- File: reproduce.py
- Name: sample_package.__file__
- Category: [ai-defects]

Sibling/follow-up issue to #666. Automatically import-provided attributes like `sample_package.__file__` are not recognized as existing by SKY-L012. 

Possible examples that should be recognized as created by import without being explicitly created include:

- `__name__`
- `__spec__`
- `__package__`
- `__loader__`
- `__file__`
- `__cached__`
- `__path__`

Other non-specifically-import-created module-object attributes skylos should potentially also recognize as existing automatically:

- `__doc__`
- `__dict__`
- `__annotations__`
- `__annotate__` (at least, for Python 3.14+)
- `__class__`

**Skylos version**
skylos 4.32.0

**Minimal reproduction**
Create sample_package/ which only needs e.g. a blank __init__.py file (e.g. [skylos_l012_implicit_module_file.zip](https://github.com/user-attachments/files/30682430/skylos_l012_implicit_module_file.zip))

```
import sample_package


package_file = sample_package.__file__

assert package_file is not None
assert package_file.endswith("__init__.py")
```
run `skylos --ai-defects . --select=SKY-L012 --format pretty`, get:

```
    █  CRITICAL  SKY-L012  Reference 'sample_package.__file__' resolves to local module 'sample_package', but '__file__' is not defined or re-exported there.
      AI Defect  reproduce.py:6
      package_file = sample_package.__file__
```

DavidPandleton:
Fixes #684

## Summary

- Recognize import-provided module attributes such as `__file__`.
- Keep `__path__` package-only and `__annotate__` Python 3.14+ only.
- Preserve detection for genuinely missing attributes.

## Tests

- 332 focused tests passed.
- 47/47 corpus cases passed.
- All GitHub checks passed.

### DIFF
--- a/skylos/rules/ai_defect/phantom_refs.py
+++ b/skylos/rules/ai_defect/phantom_refs.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 from difflib import SequenceMatcher
 from pathlib import Path
 
+from skylos.analysis.control_flow import _parse_requires_python
 from skylos.rules.vibe_dictionary import DEFAULT_VIBE_DICTIONARY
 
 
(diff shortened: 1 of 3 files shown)


mcdigman:
@duriantaco's commit was made under me while I was writing an advisory review comment, and the new commit partially addresses my concerns, but I think the rule can still be hardened further:

Other analyzers like ruff are declared-version-floor aware in some of their rules, e.g. ruff rule UP010 bans `from __future__ import annotations` if and only if pyproject.toml claims only support for `python >= 3.14` or higher, because under e.g. [PEP 749](https://peps.python.org/pep-0749/) `from __future__ import annotations` will be formally deprecated in the next few years. There are also a few similar declared-version-floor-dependent ruff rules re: e.g. exception handling. Similarly, SKY-L012 can be hardened so that `skylos` recognizes `__annotate__` as allowed without being created if and only if pyproject.toml (or I guess requirements.txt) expressly claims support for only `python >=3.14` or higher. 

Note for future reference that `from __future__ import annotations` does *not* create `__annotate__` (I had thought it would but I just verified myself that it does not in e.g. 3.12) and that no similar special handling for `__annotations__` is required because it is available for `python >= 3.0`, which is all that skylos supports.

DavidPandleton:
Thanks for the detailed review. Followed up on the version-floor point: __annotate__ is now exempted only when the scanned project declares requires-python >= 3.14 in pyproject.toml (PEP 749). Below that floor it stays a phantom reference, same as any other unknown attribute. Also confirmed your notes: from __future__ import annotations does not create __annotate__, so no extra handling there, and __annotations__ stays in the standard set since it exists since 3.0.

duriantaco (approved):

PR #709, opened 2026-08-11
fix(quality): handle positional-only defaults

### LINKED ISSUE #707: L032 Fails with nonzero exit code on positional-only arguments with defaults
**Describe the bug**
L032 fails on a function with a positional-only argument that has a default. The issue appears to be that _iter_arg_defaults in skylos/quality/logic_security.py does not index positional-only arguments. 

Since 4.31.1, the rule fails closed, while on 4.31.0 and prior, it still failed, but failed open; the difference is just that PR #654 ensured incomplete analysis causes skylos to fail closed; the behavior was not introduced by 4.31.1.

Note: the reproducer below is distilled from [psf/src/requests/_types.py](https://github.com/psf/requests/blob/main/src/requests/_types.py), and psf/requests is a documented skylos `quality` benchmark source already (see e.g. skylos/benchmarks/quality/manifest.json), so it is within-scope to mandate good behavior on it.

PEP 484 creates support for type hints with an ellipsis-default  `...` parameter, and PEP 570 creates support for positional-only parameters, so the original and reproducer are syntactically legal type hints that should be supported. The original is a legitimate PEP 544 runtime_checkable Protocol, so L032 should not actually *fire* in this case, just complete its analysis.

**To Reproduce**

```
from typing import Protocol


class SupportsRead(Protocol):
    """Minimal protocol containing a defaulted positional-only parameter."""

    def read(self, length: int = ..., /) -> bytes:
        """Read up to ``length`` bytes."""
        ...
```

Skylos fails with an error code with:

    █  HIGH  SKY-ANALYSIS-INCOMPLETE  list index out of range
      Analysis  reproduce.py:1
      """Reproduce Skylos quality analysis failing on a positional-only default."""
[skylos_l032_positional_only_default.zip](https://github.com/user-attachments/files/30951175/skylos_l032_positional_only_default.zip)

**Expected behavior**
No error; L032 should not actually *fire* on this example, because it is a PEP 544 Protocol

**Actual behavior**
Analysis incomplete error.

**Environment**
- Python version: 3.14.6

mcdigman:
Closes #707.

Note: Although not mentioned in the original issue, SKY-L014 is also affected by the fix, because it reuses the same helper.

## Summary

- map positional defaults across both `posonlyargs` and ordinary `args`
- prevent crashes for all-positional-only defaults and preserve names for mixed positional defaults
- verify SKY-L032 fires for placeholder string defaults on both Protocol declarations and concrete implementations
- retain SKY-L014 coverage for the shared default-argument helper

## Testing

- `pytest -q test/test_quality_rules.py test/test_analyzer.py` — 214 passed
- exact Protocol `length: int = ...` reproducer — no analysis errors and no SKY-L032 finding
- Protocol `"test@example.com"` positional-only default — SKY-L032 reported with the correct parameter name
- full suite — 6030 passed, 7 skipped, with 2 unrelated failures that reproduce on pristine `main` in the local Python 3.14 environment

## AI Use Disclosure
Codex GPT-5.6 Sol drafted the PR, and it, including the testing coverage, was reviewed per repository llm review guidelines by both Sol and Claude Opus-5.0. The analyzer changes are just 3 lines and human verified as well.

### DIFF
--- a/skylos/rules/quality/logic_security.py
+++ b/skylos/rules/quality/logic_security.py
@@ -1338,12 +1338,13 @@ class MockPlaceholderDataRule(SkylosRule):
 def _iter_arg_defaults(func_node):
     args = func_node.args
     num_defaults = len(args.defaults)
-    num_args = len(args.args)
+    positional_args = [*args.posonlyargs, *args.args]
+    num_args = len(positional_args)
     offset = num_args - num_defaults
 
     for i, default in enumerate(args.defaults):
         if default:
-            yield args.args[offset + i], default
+            yield positional_args[offset + i], default
     for arg, default in zip(args.kwonlyargs, args.kw_defaults):
         if default:
             yield arg, default
--- a/test/test_analyzer.py
+++ b/test/test_analyzer.py
@@ -3388,6 +3388,50 @@ def fake_call():
             "low_entropy_uuid",
         }
 
+    def test_analyze_flags_protocol_and_implementation_defaults(
+        self,
+        tmp_path,
+        monkeypatch,
+    ):
+        monkeypatch.setenv("SKYLOS_JOBS", "1")
+        src = tmp_path / "protocols.py"
+        src.write_text(
+            """
+from typing import Protocol
+
+class Notifier(Protocol):
+    def notify(
+        self,
+        protocol_email: str = "test@example.com",
+        /,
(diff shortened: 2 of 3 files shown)


duriantaco (approved):
LGTM

PR #712, opened 2026-08-12
fix(security): handle positional-only MCP defaults

### LINKED ISSUE #710: SKY-D244 Fails on positional-only arguments with defaults
**Describe the bug**
Sibling to #707 for SKY-D244; _check_param_defaults in skylos/rules/danger/danger_mcp/mcp_flow.py has an _identical_ flaw to the one I propose a fix to in #709; the identical patch to the function (plus unit tests) should fix it. 

**To Reproduce**
Reproducer attached that covers several edge cases that can cause the secret to be either missed or misattributed to the wrong variable. 

[skylos_d244_positional_only_secret.zip](https://github.com/user-attachments/files/30963513/skylos_d244_positional_only_secret.zip)

One critical example:

```
from mcp.server.fastmcp import FastMCP


mcp = FastMCP("demo")


@mcp.tool()
def fetch(url: str, api_key: str = "sk-AbCd1234EfGh5678IjKlMnOp", /, note: str = "hi") -> str:
    """The hardcoded secret is in ``api_key``; ``note`` holds a harmless string."""
    return url + api_key + note
```

returns

```
  misattributed.py · 1 issue

    █  CRITICAL  SKY-D244  Hardcoded secret in MCP tool parameter default 'note'.
      Security  misattributed.py:10
      def fetch(url: str, api_key: str = "sk-AbCd1234EfGh5678IjKlMnOp", /, note: str = "hi") -> str:
````

which incorrectly attributes the secret to the variable 'note' instead of 'api_key'. 

**Expected behavior**
Expected behavior for various cases described in README of reproducer; 

**Actual behavior**
Described in README and case above

**Environment**
- Python version: 3.14.6

mcdigman:
Closes #710.

Note: This is the SKY-D244 sibling to #709. The MCP security scanner carries an independent copy of the positional-default alignment logic, so the code changes do not overlap.

## Summary

- map positional defaults across both `posonlyargs` and ordinary `args`
- prevent all-positional-only defaults from aborting the MCP scan and suppressing later findings
- preserve the correct parameter name for secrets in mixed positional signatures
- add regression coverage based on the issue reproducers

## Testing

- `pytest -q test/test_mcp_rules.py -k 'D244'` — 8 passed
- `pytest -q test/test_mcp_rules.py` — 34 passed
- all-positional-only reproducer — both `api_key` and the later `token` reported as SKY-D244
- mixed positional reproducer — SKY-D244 attributed to `api_key`, not the harmless `note`
- ordering reproducer — all three secrets reported regardless of declaration order
- full suite — 6027 passed, 7 skipped, with 2 unrelated failures that reproduce on pristine `main` in the local Python 3.14 environment

## AI Use Disclosure
Codex GPT-5.6 Sol drafted the PR, and it, including the testing coverage, was reviewed per repository llm review guidelines by both Sol and Claude Opus-5.0. The analyzer changes are just 3 lines and human verified as well.


duriantaco (approved):
Reviewed and tested. The fix is correct, and all CI checks pass.

PR #715, opened 2026-08-12
fix: fail closed when analysis subsystems raise instead

### LINKED ISSUE #713: SKY-D24X currently fails open
**Describe the bug**
Issue #710 wasn't surfaced earlier because a blind `except Exception`  in mcp_flow.py trapped it and wiped out _all_ SKY-D24X reports, causing it to fail open:
```
    except Exception as e:
        print(f"MCP analysis failed for {file_path}: {e}", file=sys.stderr)
```

It should fail closed instead/not wipe the SKY-24X scan results.

Sibling to #692.

Note the trigger the attached reproducer uses will be silenced by #712, because it is the only precisely analogous trigger mode I have presently identified; the other possible trigger (that wouldn't be fixed by #712) is that a very deep nested recursion (>~500) can also trigger the SKY-24X scan results to be wiped, but that case already fails closed (exit code 2) for a different reason, so it isn't a _perfect_ reproducer because it at least the failure isn't silent.

A robust fix should fix both.

**To Reproduce**
See #710 or the attached reproducer demonstrating that scan errors disappear. 

[skylos_mcp_scan_swallows_errors.zip](https://github.com/user-attachments/files/30965816/skylos_mcp_scan_swallows_errors.zip)

**Expected behavior**
Fail closed/successful SKY-24X results not swallowed.

**Actual behavior**
Fails open

**Environment**
- Python version: 3.14.6

mcdigman:
## Relevant Issues
Closes #713, plus applies essentially the same fix to 15+ additional sites, to harden skylos against potentially similar failure classes in the future. Subject to several known limitations documented below, which would either be scoped to a separate PR or require a design/API decision (such as whether to clean up the possibly dead `danger.py:590 scan_ctx`). 

## What does this PR do?

Makes analysis failures visible instead of silent. Skylos caught exceptions in three places and continued with only a log line, so a scanner that crashed produced the same output as one that found nothing: exit 0, empty `analysis_errors`, no indication that part of the scan never ran.

Three families are addressed:

1. **Taint scanners** — all 16 `scan()` functions in `skylos/rules/danger/danger_*/` caught `Exception` and printed one line to stderr.
2. **The taint dispatcher** — `analyzer.py` wrapped `scan_file_with_tree` in `logger.debug`, which is invisible at default verbosity and also covers `_PythonCommandExfilChecker`.
3. **`analyze()` feature blocks and worker extraction** — 15 subsystem blocks (secrets, prompt injection, dependency/SCA, AI-defect, repo policy, security regressions) plus clone extraction and architecture metrics in `proc_file`.

Each now records a `SKY-ANALYSIS-INCOMPLETE` entry naming what failed, routed into `analysis_errors`. No new exit-code machinery: `_analysis_incomplete_exit_code` already returns 2 for a non-empty `analysis_errors`.

This follows the pattern already established for incomplete analysis by the container false-clean fix and the Go engine-unavailable fix, which introduced `SKY-ANALYSIS-INCOMPLETE` for exactly this situation.

## Behavior changes

- **Exit code.** A run where a scanner or subsystem raises now exits **2** instead of 0. CI that treats non-zero as failure will start failing on repositories where analysis is partially broken. This is the point of the change, but it is the item most likely to surprise.
- **`analysis_errors` gains entries** with `rule_id: SKY-ANALYSIS-INCOMPLETE`, `severity: HIGH`, `kind: processing_error`, plus a `scanner` (taint) or `subsystem` (analyze/worker) field naming the component.
- **stderr messages removed.** The 16 `print(f"... analysis failed for {path}: {e}", file=sys.stderr)` lines are replaced by structured findings. Anything grepping stderr for those strings needs updating. One existing test did exactly that (`test_xss_flow.py`) and has been rewritten to assert the recorded payload.
- **`mcp_flow.scan()` gate moved inside its `try`.** It evaluated `_is_mcp_file(tree)` outside the handler, so a failure there escaped `scan()` entirely rather than being recorded.
- **`analysis_error_count`** is recomputed immediately before no-source results are returned; optional subsystems append after the earlier count was taken, so it could disagree with `len(analysis_errors)`.
- **Per-file secret failures name the file**, not the scan root, so GitHub annotations point at a real path.

## Testing

New coverage is deliberately structural rather than case-by-case, so a component added later that swallows its failures fails these tests:

- `test/test_scanner_failure_reporting.py` — enumerates the taint scanners via `pkgutil` (16 discovered) and forces each into its failure path; also asserts each `scan()` body is fully guarded, and that the incomplete marker lands in `analysis_errors` rather than the findings list.
- `test/test_subsystem_failure_reporting.py` — extracts `analyze()`'s broad handlers by AST (22) and `proc_file`'s (4), asserting each records rather than only logging; plus end-to-end cases for secret scanning (`.py` and `.env` paths), per-file attribution, and the summary count.

Suite result: **6077 passed**. Two failures are pre-existing and reproduce on clean `main` (`test_observations_reject_excessive_json_nesting_without_crashing` and its OpenAI twin).

## Precision Impact

No rule logic changes; no finding is added, removed, or reclassified for code that analyzes successfully. The only new output is `SKY-ANALYSIS-INCOMPLETE`, emitted only when a component raises.

## Known limitations and leftovers

Deliberately out of scope, each worth its own issue:

- **`danger.py:590 scan_ctx`** is a second entrypoint with its own swallow. It appears unused (only `scan_file_with_tree` is imported anywhere), so it was left alone rather than guessing at a plugin contract.
- **Local API hallucination detectors** (Python/Go/Java/JS) record `failed_*_api_check` and set `ai_verification.state: incomplete`, but the exit-code rule ignores that field, so those failures still exit 0. Gating it is a policy decision: a missing Go or Java toolchain reaches the same path, so ordinary machines would start exiting 2.
- **Git change auto-detection** still swallows. On failure `changed_files` stays `None`, which is also the legitimate "no changes" state, so recording an error there changes the meaning of a normal outcome.
- **The phantom fallback re-reads and re-parses inside its loop**, so one unreadable file drops phantom detection for every remaining file. Fixing that is a behavioral change beyond error reporting.
- **Custom and community rule loading** (`analysis/file_processing.py`) still fails open — a malformed rule file means those rules silently never run.
- `_incomplete.py` duplicates the payload shape instead of sharing `_analysis_error_payload`, which lives in `analyzer.py` and cannot be imported from a rules module without a cycle. `python_version` is kept in step by hand and asserted in tests.
- The structural tests match handlers by call-name markers, so a subsystem whose call names avoid `scan`/`detect`/`analyze`/`rule`/`visitor`/`linter` could escape the guard. The `analyze()` count is pinned at its real value so coverage cannot shrink unnoticed.

Scope of the underlying risk: apart from `mcp_flow`, these handlers are **latent** rather than actively firing. A search of 432 variants across all 16 scanners — each engaged with a fixture that produces a real finding, crossed with modern-syntax and signature mutations — found no other live trigger, with a positive control confirming the search detects the known `mcp_flow` bug.

## Checklist

- [x] Tests pass (`python3 -m pytest test/`)
- [x] No new false positives introduced (if modifying analysis logic)

## AI Use Disclosure

The sites to change were identified by a sweep with Claude Opus 5.0, as were the respective patches. Code review was done per skylos llm agent instructions using both GPT-5.6 Sol and a separate Claude Opus 5.0 agent. I reviewed all the changes by hand; they are largely essentially the same change repeated near-identically at every relevant site, plus unit tests.

### DIFF
--- a/skylos/analyzer.py
+++ b/skylos/analyzer.py
@@ -4656,7 +4656,10 @@ def proc_file(
                         if line.strip() and not line.strip().startswith("#")
                     ),
                 }
-            except Exception:
+            except Exception as exc:
+                scanner_errors.append(
+                    _subsystem_error_payload(file, "Architecture metrics", exc)
+                )
                 logger.debug(
                     "Architecture metric extraction failed for %s",
                     file,
@@ -4674,7 +4677,10 @@ def proc_file(
                 clone_fragments = extract_fragments(
                     Path(file), source, clone_cfg, tree=clone_tree
                 )
-            except Exception:
+            except Exception as exc:
+                scanner_errors.append(
+                    _subsystem_error_payload(file, "Clone extraction", exc)
+                )
                 logger.debug(
                     "Clone fragment extraction failed for %s", file, exc_info=True
                 )
--- a/skylos/analyzer.py
+++ b/skylos/analyzer.py
@@ -2701,7 +2701,12 @@ class Skylos:
                             if finding.get("rule_id") in project_ignore:
                                 continue
                             ai_defect_findings.append(finding)
-                    except Exception:
+                    except Exception as exc:
+                        result["analysis_errors"].append(
+                            _subsystem_error_payload(
+                                path, "Manifest dependency hallucination scan", exc
+                            )
+                        )
                         if os.getenv("SKYLOS_DEBUG"):
                             logger.error(
                                 "Manifest dependency scan failed", exc_info=True
@@ -2791,6 +2796,9 @@ class Skylos:
                         result["dependency_vulnerabilities"] = list(sca_findings)
                         result["analysis_summary"]["sca_count"] = len(sca_findings)
                     except Exception as exc:
+                        result["analysis_errors"].append(
+                            _subsystem_error_payload(path, "Dependency scan", exc)
+                        )
                         self._sca_coverage = {
                             "status": "incomplete",
                             "complete": False,
@@ -3164,7 +3172,10 @@ class Skylos:
                                         )
                                     ]
                                 all_secrets.extend(findings)
-                        except Exception:
+                        except Exception as exc:
+                            analysis_errors.append(
+                                _subsystem_error_payload(path, "Secret scan", exc)
+                            )
                             logger.debug("Secret scan failed for file", exc_info=True)
 
             if enable_secrets and _secrets_scan_ctx is not None:
@@ -3319,7 +3330,10 @@ class Skylos:
                             cf,
                         )
                         all_quality.extend(reg_findings)
-            except Exception:
+            except Exception as exc:
+                analysis_errors.append(
+                    _subsystem_error_payload(path, "Security regression detection", exc)
+                )
                 if os.getenv("SKYLOS_DEBUG"):
                     logger.error("Security regression scan failed", exc_info=True)
 
@@ -3334,7 +3348,12 @@ class Skylos:
                         changed_files=changed_files,
                     )
                 )
-            except Exception:
+            except Exception as exc:
+                analysis_errors.append(
+                    _subsystem_error_payload(
+                        path, "Security contract regression detection", exc
+                    )
+                )
                 if os.getenv("SKYLOS_DEBUG"):
                     logger.error("Security contract scan failed", exc_info=True)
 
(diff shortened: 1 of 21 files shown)


duriantaco:
Thanks @mcdigman for reporting this! We resolved #713 with a fix updated for the current analyzer architecture, including the deep recursion case. imo the fail-closed hardening in this pr is still valuable, but should be rebased and split into a separate follow up pr if possible. Thanks once again!

PR #721, opened 2026-08-14
perf(analyzer): batch grep verification searches

### LINKED ISSUE #722: Speed up grep verification
Grep verification is currently the limiting factor on skylos runtime. Because it is slow, #714 (non-determinism due to cut offs) currently bites harder than it should, and creates a lot of false positives on large repositories that would have been recovered by the grep verification.

Because the `liveness_primer` CI proposed in #720 needs to run the full deterministic scan with no cutoff (except for sanity/CI minutes), the slowness bites it, and the feasibility of using it as a CI, particularly hard. 

**Proposed solution**
By manual inspection/cProfile, I determined that the main limiting factor is that calls to rg are not batched, so the entire tree of the target repository is grepped a *huge* number of times; sometimes 10s of thousands, in the test I did averaging 6.7 traversals per finding. 

The most obvious solution is batching the calls to `rg`, which I implemented in PR #721. Performance improvements are documented in the PR  body, but some details are reiterated here in case the PR is superseded.

I kept several paths like the `rg` not installed fallback and parallel implementation that *don't* batch the commands, which helps with verifying the behavior has not changed. 

In addition to potentially facilitating `liveness_primer` running as a CI in a reasonable number of CI minutes, this change can be verified by `liveness_primer` itself; I didn't add the proposed workflow, but when run manually, it verifies the refactor carries 0 changed findings in the entire 15 project corpus, while `liveness_primer` itself runs several times faster. 

Note while the PR makes #714 not bite quite as hard, it does not resolve it; I left that to a separate commit, and in fact the current commit deliberately works around the existence of the hard walltime cutoff in its batching strategy; if the walltime cutoff was altered as suggested in #714, some of that code could potentially be stripped out. 

Also, I left a few issues to maintainer discretion; the non-batched parallel path could currently be delete entirely, because it is now much slower than the batched serial path. The python fallback path *could* be batched also, although the performance improvement would be far more limited. 

Incidentally, the batching greatly reduces the performance pressure on the caching implementation so the proposed wiping of the grep cache for detected secrets in #706 wouldn't bite as hard.

**Alternatives considered**
This workflow could be offloaded *entirely* into rust/rg, instead of batched, and might be somewhat faster if done that way, but that would be a larger change to the codebase for probably limited marginal performance improvement. The batching of 128 is somewhat arbitrary, more batching would probably be a larger speedup but more danger of the #714 walltime cutoff biting if it isn't overridden.

mcdigman:
Closes #722

## What this does

Grep verification was repeatedly traversing the repository for individual findings. This PR records compatible requests, executes them in deterministic batches with `rg`, and replays the results through Skylos's existing strategy and verdict logic.

The original contributor benchmark measured:

| Scope | `main` | batched | speedup |
| --- | ---: | ---: | ---: |
| 15 pinned corpus projects | 265.66 s | 57.31 s | 4.64x |
| mitmproxy | 88.29 s | 14.10 s | 6.26x |

The same corpus comparison reported 819 findings on both sides, with no added, dropped, or changed findings.

## Maintainer hardening

- preserves paths, line numbers, and content using ripgrep JSON or NUL-delimited grep output, including Windows and colon-containing paths
- keeps Python and ripgrep regex behavior aligned for Unicode boundaries and engine-sensitive patterns
- applies one shared deadline and bounded input, output, match-count, and cleanup limits
- isolates noisy batches without turning a timed-out batch into unbounded serial work
- rejects grep executables resolved from the scanned repository, single-file scan directory, or current working directory
- keeps verification caching bounded and process-local; repository-owned cache files are not trusted, read, or written
- preserves structured evidence through LLM verification and TypeScript/Python override paths

All six contributor commits remain in the branch, followed by two focused maintainer hardening commits.

## Validation

- `pytest -q -W error::ResourceWarning test/test_grep_cache.py test/test_grep_verify.py test/test_verify_orchestrator.py` — 283 passed
- analyzer grep/cache regression set — 15 passed
- security benchmark — 67 cases, 0 failures, 100/100
- corpus guard — 47 cases, 0 failures
- Skylos self-scan — 0 danger findings and 0 analysis errors
- staged security, secrets, quality, and AI-defect gate — passed
- full local suite — 6,899 passed and 15 skipped; remaining failures were confined to optional-language parser/benchmark areas, with no failures in the changed grep/cache/LLM suites

## Attribution

Thanks @mcdigman for the performance investigation, batching design, implementation, benchmarks, and follow-up fixes. The contributor history and attribution are preserved.

## AI use disclosure

The contributor disclosed Codex GPT-5.6 Sol and Claude Fable 5 assistance for the original implementation and review. Maintainer hardening was implemented and independently reviewed with Codex agents under the repository's agent guidelines.

### DIFF
--- a/skylos/core/grep_verify.py
+++ b/skylos/core/grep_verify.py
@@ -10,10 +10,13 @@ from typing import Any
 
 from skylos.core.grep_verify_common import (
     GrepRequest,
+    _GrepEvidence,
+    _GrepExecutionIncomplete,
     _run_grep,
     detect_language,
     execute_grep_batch,
     filter_grep_results,
+    grep_execution_deadline,
     is_definition_line,
     is_substring_match,
     module_candidates,
@@ -99,10 +102,11 @@ class _PendingBatchFinding:
     requests: tuple[GrepRequest, ...]
 
 
-_GREP_VERIFY_CACHE_VERSION = "v6"
-# The deadline is checked between groups. This bounds overshoot while ensuring
-# a started group produces complete, replayable findings instead of none.
+_GREP_VERIFY_CACHE_VERSION = "v8"
 _GREP_FINDING_BATCH_SIZE = 32
+_GREP_CACHE_MAX_STRATEGIES = 64
+_GREP_CACHE_MAX_LINES_PER_STRATEGY = 256
+_GREP_CACHE_MAX_EVIDENCE_CHARS = 1_000_000
 
 _DETERMINISTIC_RULES: list[tuple[str, str, str]] = [
     ("method_calls", "real_method_call", "Direct method-call usage found via grep"),
@@ -160,10 +164,38 @@ def _load_cached_group_results(
     if cached is None:
         return None
     try:
-        return _json.loads(cached[0]) if cached else {}
+        decoded = _json.loads(cached[0]) if cached else {}
     except (_json.JSONDecodeError, TypeError, ValueError) as exc:
         logger.debug("Ignoring invalid grep verification cache entry: %s", exc)
         return None
+
+    return _normalize_cached_group_results(decoded)
+
+
+def _normalize_cached_group_results(
+    decoded: object,
+) -> dict[str, list[str]] | None:
+    if not isinstance(decoded, dict) or len(decoded) > _GREP_CACHE_MAX_STRATEGIES:
+        logger.debug("Ignoring invalid grep verification cache result shape")
+        return None
+
+    normalized: dict[str, list[str]] = {}
+    evidence_chars = 0
+    for strategy, lines in decoded.items():
+        if (
+            not isinstance(strategy, str)
+            or not isinstance(lines, list)
+            or len(lines) > _GREP_CACHE_MAX_LINES_PER_STRATEGY
+            or not all(isinstance(line, str) for line in lines)
+        ):
+            logger.debug("Ignoring invalid grep verification cache strategy")
+            return None
+        evidence_chars += sum(len(line) for line in lines)
+        if evidence_chars > _GREP_CACHE_MAX_EVIDENCE_CHARS:
+            logger.debug("Ignoring oversized grep verification cache evidence")
+            return None
+        normalized[strategy] = lines
+    return normalized
 
 
 def _store_cached_group_results(
@@ -172,11 +204,21 @@ def _store_cached_group_results(
     finding: dict,
     results: dict[str, list[str]],
 ) -> None:
+    normalized = _normalize_cached_group_results(results)
+    if normalized is None:
+        return
+    if any(
+        isinstance(line, _GrepEvidence) and ":" in line.path
+        for lines in normalized.values()
+        for line in lines
+    ):
+        logger.debug("Skipping cache entry with an ambiguous legacy evidence path")
+        return
     cache_key = _cache_key(cache, group_name, finding)
     if cache_key is None:
         return
     try:
-        cache.put(cache_key, [_json.dumps(results)])
+        cache.put(cache_key, [_json.dumps(normalized)])
     except (AttributeError, OSError, TypeError, ValueError) as exc:
         logger.debug("Failed to write grep verification cache entry: %s", exc)
 
@@ -305,16 +347,31 @@ def _execute_pending_findings(
     pending: list[_PendingBatchFinding],
     project_root: str,
     cache: Any,
+    deadline: float,
 ) -> dict[str, GrepVerdict]:
     requests = [request for item in pending for request in item.requests]
-    batch_results = execute_grep_batch(requests)
+    batch_results = execute_grep_batch(requests, deadline=deadline)
+    incomplete_requests = getattr(batch_results, "incomplete_requests", set())
     verdicts: dict[str, GrepVerdict] = {}
     for item in pending:
-        with replay_grep_results(batch_results):
-            search_results = multi_strategy_search(item.finding, project_root)
-        _store_cached_group_results(
-            cache, item.group_name, item.finding, search_results
-        )
+        if time.monotonic() >= deadline:
+            break
+        if any(request not in batch_results for request in item.requests):
+            continue
+        try:
+            with replay_grep_results(batch_results, deadline=deadline):
+                search_results = multi_strategy_search(item.finding, project_root)
+        except _GrepExecutionIncomplete as exc:
+            logger.debug("grep replay was incomplete: %s", exc)
+            continue
+        if time.monotonic() >= deadline:
+            break
+        if not any(
+            request in incomplete_requests for request in item.requests
+        ):
+            _store_cached_group_results(
+                cache, item.group_name, item.finding, search_results
+            )
         verdict = _apply_deterministic_rules(search_results, item.finding)
         if verdict:
             verdicts[_finding_full_name(item.finding)] = verdict
@@ -333,7 +390,7 @@ def _grep_verify_findings_batched(
     deadline = start_time + time_budget
 
     for finding in findings:
-        if time.monotonic() > deadline:
+        if time.monotonic() >= deadline:
             break
         full_name = _finding_full_name(finding)
         if not full_name:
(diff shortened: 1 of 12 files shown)


skylos/core/grep_verify.py
```diff
@@ -106,37 +121,64 @@ def _cached_group_results(
     finding: dict,
     search_fn: Callable[[], dict[str, list[str]]],
 ) -> dict[str, list[str]]:
-    if cache is None or not group_name:
-        return search_fn()
+    cached = _load_cached_group_results(cache, group_name, finding)
```
mcdigman, line 128: This change is basically just a refactor that reduces cognitive complexity, but it is not just churn/ is needed because e.g. `_plan_batched_finding` and `_execute_pending_findings` require access to the separate `_load_cached_group_results` and `_store_cached_group_results`. The behavior of `_cached_group_results` itself is functionally identical (except theoretically for a race condition where the target repo changes underneath it *while the function is executing*, in which case the behavior cannot be guaranteed to be strictly identica to the prior behavior)

skylos/core/grep_verify.py
```diff
@@ -230,6 +272,182 @@ def _apply_deterministic_rules(
...
+    if deterministic_verdict:
+        return deterministic_verdict, None
+
+    group_name = _serial_group_name(finding)
+    cached = _load_cached_group_results(cache, group_name, finding)
+    if cached is not None:
+        return _apply_deterministic_rules(cached, finding), None
+
+    with record_grep_requests() as recorded:
+        planned_results = multi_strategy_search(finding, project_root)
```
mcdigman, line 337: Untouched logic

skylos/core/grep_verify.py
```diff
@@ -230,6 +272,182 @@ def _apply_deterministic_rules(
...
+    cache: Any,
+) -> tuple[GrepVerdict | None, _PendingBatchFinding | None]:
+    deterministic_verdict = _deterministic_suppression_verdict(finding)
+    if deterministic_verdict:
+        return deterministic_verdict, None
+
+    group_name = _serial_group_name(finding)
+    cached = _load_cached_group_results(cache, group_name, finding)
+    if cached is not None:
+        return _apply_deterministic_rules(cached, finding), None
```
mcdigman, line 334: Untouched logic

skylos/core/grep_verify.py
```diff
@@ -230,6 +272,182 @@ def _apply_deterministic_rules(
...
+    language = _finding_language(finding)
+    return "python_core" if language == "python" else f"serial_{language}"
+
+
+def _plan_batched_finding(
+    finding: dict,
+    project_root: str,
+    cache: Any,
+) -> tuple[GrepVerdict | None, _PendingBatchFinding | None]:
+    deterministic_verdict = _deterministic_suppression_verdict(finding)
```
mcdigman, line 327: Untouched logic

skylos/core/grep_verify.py
```diff
@@ -230,6 +272,182 @@ def _apply_deterministic_rules(
...
+    verdicts: dict[str, GrepVerdict] = {}
+    for item in pending:
+        with replay_grep_results(batch_results):
+            search_results = multi_strategy_search(item.finding, project_root)
+        _store_cached_group_results(
+            cache, item.group_name, item.finding, search_results
+        )
+        verdict = _apply_deterministic_rules(search_results, item.finding)
+        if verdict:
+            verdicts[_finding_full_name(item.finding)] = verdict
```
mcdigman, line 377: Untouched logic

skylos/core/grep_verify.py
```diff
@@ -230,6 +272,182 @@ def _apply_deterministic_rules(
...
+                )
+    finally:
+        for future in pending:
+            future.cancel()
+        executor.shutdown(wait=True, cancel_futures=True)
+
+    return verdicts
+
+
 def grep_verify_findings(
```
mcdigman, line 518: This function already had high cognitive complexity (~40, and cyclomatic complexity ~20) and the additional complexity from the batched processing would have put it over the edge to near-unreadability, so refactor splits out the execution into helpers.

skylos/core/grep_verify.py
```diff
@@ -239,93 +457,24 @@ def grep_verify_findings(
...
-    for finding in remaining_findings:
-        if time.monotonic() - start_time > time_budget:
-            break
-
-        full_name, verdict = process_finding(finding)
-        if verdict:
-            verdicts[full_name] = verdict
-
-    return verdicts
+    return _grep_verify_findings_parallel(
```
mcdigman, line 542: The unchanged parallel branch is now much _slower_ than the batched branch, even with ~8 workers; it is extracted/separated to reduce complexity but kept for compatibility; consider removing it entirely, which would reduce the apparent complexity/potentially increase maintainability of this functionality. It might theoretically be possible to parallelize the batched execution, but doing so might high disk/RAM bottlenecks before it realized any practical advantage.

skylos/core/grep_verify_common.py
```diff
@@ -108,78 +156,374 @@ def source_globs_for_language(lang: str) -> list[str]:
...
+    filtered: list[str] = []
+    for line in stdout.strip().splitlines():
+        normalized = line.replace("\\", "/")
+        if any(part in normalized for part in _IGNORED_GREP_PATH_PARTS):
+            continue
+        filtered.append(line)
+    return filtered
 
+
+def _run_grep_request(request: GrepRequest) -> list[str]:
```
mcdigman: This function is broken up partly for complexity, and partly to facilitate the batching

skylos/core/grep_verify_common.py
```diff
@@ -108,78 +156,374 @@ def source_globs_for_language(lang: str) -> list[str]:
...
+
+def _run_ripgrep_batch(
+    requests: Sequence[GrepRequest],
+    rg: str,
+    timeout: float = 30.0,
+) -> dict[GrepRequest, tuple[str, ...]]:
+    batched, direct = _partition_grep_requests(requests)
+    batch_results: dict[GrepRequest, tuple[str, ...]] = {}
+    if not batched:
+        return _run_serial_grep_requests(direct)
```
mcdigman: Serial fallback is necessary if e.g. 1 of the 128 in the batch fail, so the other 127 can still run. This could result in a drastic slowdown though especially if the batch size is made bigger; technically there are alternatives like recursive bisection to identify the failing expression, but any performance improvement might not justify the added complexity. It probably would be noisy to fail with a nonzero exit code in this case instead.

skylos/core/grep_verify_common.py
```diff
@@ -108,78 +156,374 @@ def source_globs_for_language(lang: str) -> list[str]:
...
-            for d in _GREP_EXCLUDE_DIRS:
-                if any(ch in d for ch in "*?["):
-                    cmd.extend(["-g", f"!**/{d}/**"])
-                else:
-                    cmd.extend(["-g", f"!**/{d}/**"])
-            cmd.extend(["--", pattern, project_root])
+            cmd = _ripgrep_command(request, rg)
+            cmd.extend(["--", request.pattern, request.project_root])
         else:
             grep_flags = ["-rn"]
```
mcdigman: We could also batch the rg not present case, but rg probably usually is present when speed is desired and rg not present will be slower anyway; follow up commit if desired. For now, preserving the fallback logic is a secondary way of showing that the batching is behavior preserving.

mcdigman (commented):
Batching is separated as expected

skylos/core/grep_verify_common.py
```diff
@@ -108,78 +156,374 @@ def source_globs_for_language(lang: str) -> list[str]:
...
+        timeout=timeout,
+        check=False,
+    )
+    if result.returncode not in (0, 1):
+        msg = result.stderr.strip() or f"exit status {result.returncode}"
+        raise RuntimeError(msg)
+    return sorted(_filter_grep_output(result.stdout), key=_grep_line_sort_key)
+
+
+def _ripgrep_stdin_match_positions(
```
mcdigman, line 945: Additional functionality to better guarantee _identical_ behavior to baseline/no surprises with non-ascii text; not needed if minor deviations are allowable; this commit doesn't move any diffs on the liveness_primer corpus either way

duriantaco (approved):
We kept your commits and added maintainer follow ups for safety and edge cases. All CI checks pass. Approved and ready to merge.

skylos/core/grep_verify.py
duriantaco: Kept for compatibility in this PR. We can consider removing it separately after the batched path has had prod use.

skylos/core/grep_verify_common.py
duriantaco: Addressed in the maintainer follow up: failing batches are now split to isolate problematic requests, timeouts do not trigger serial fan out.

PR #732, opened 2026-08-16
ci(analyzer): add real-world blast radius check

### LINKED ISSUE #720: Integrate liveness_primer into CI
Both large refactors and subtle changes can have unintended consequences on external repositories that use skylos. This can result in missed false positives, dropped findings, or bugs that have to be fixed later, missed behavioral drift due to refactors, and wasted time *verifying* a refactor has *not* caused behavioral drift by code analysis/inspection rather than simply using automated empirical validation on a larger external corpus or targeted repositories of interest.

To help address this problem, I build [`liveness_primer`](https://github.com/mcdigman/liveness_primer). It is based on the similar tool `mypy_primer` for type checkers; it does identical parallel runs of the base and head commits of skylos against a larger external corpus of open-source repositories at either pinned commits or latest, currently mostly seeded with existing skylos corpus sources, to identify empirically the *blast radius* that a PR to skylos will have. 

liveness_primer's behavior is different from what corpus/README.md currently flags as justification for not scanning whole repos:
```
Why this is better than scanning whole upstream repositories on every PR:

Whole-repo scans are noisy because mature libraries may contain legitimate findings unrelated to the regression you are trying to catch.
Whole-repo scans are slow and brittle because upstream code changes over time.
Distilled fixtures let us assert exact expectations like "this symbol must not be reported as dead code" without hiding unrelated issues.
```

by scanning it before and after and isolating only the *diff*, we reduce noise of both legitimate findings and existing false-positives unrelated to the current change under test in the PR. Brittleness due to upstream code changes is resolved by running the corpus only against pinned commits; liveness_primer can always also manually be run against the upstream head from the command line, and the pinned commits can always be updated. It is even possible to run multiple pinned commits from the *same* upstream repository, if large upstream refactors make multiple commits interesting regression guards for skylos. 

Because `liveness_primer` measures the *blast radius* for human or llm inspection, the CI doesn't fail unless skylos exits with a non-zero exit code on an upstream repository, in which case the failure is objective and strictly a high-value finding. We don't have to assert exact expectations, although it would be possible to add heuristic or llm-based triage features in the future. 

In addition to a CI, and text and markdown outputs, the full json reports can be explored in human-readable form with a static report explorer tool I built, screenshotted below. It is possible to have a *separate* CI workflow summarize the results in a PR comment, either with automatic extraction from the markdown report or with an llm pre-prompted to summarize the diffs in the from the JSON (the commenting workflow must be separate for trust-boundary reasons; the main workflow cannot be allowed to have write access, because a new PR is attacker-controlled code and `liveness_primer` runs it in a no-network sandbox). Examples of both cases *on a real skylos release version* are described in the liveness_primer README and screenshotted below.

As for the cost in CI minutes, the current chosen pins are chosen to be roughly in line with the current tests workflows; if using it as a CI provides material value, this cost is not unreasonable. I will add some examples of cases where the CI minutes would earn their keep in follow-up issues. One such case for skylos is already mentioned in the README. 

Note that many of the issues I have recently flagged have been identified either by test runs of `liveness_primer` or running skylos on `liveness_primer` in the process of building it, so the tool has already materially contributed to improving skylos.

The tool is currently the first release version, and I would appreciate any feedback and would be happy to make any changes that would make it more useful to you.

Note that although the name features `liveness`, it is capable of running all of the skylos rulesets. For the initial demonstration, all the corpus sources only run with dead code detection, but the other rules can easily be added to the defaults or a CI-specific corpus, or manually run from the `liveness_primer` command line.

**Screenshots of potential use cases***
Image
Image
Image

duriantaco:
Closes #720

## Summary

  - add a read-only PR workflow that compares the base revision against the proposed merge
  - run both Skylos revisions across 15 pinned real-world Python projects
  - show new, dropped and changed findings without blocking intentional analyzer changes
  - fail on crashes, timeouts, incomplete analysis or invalid reports

  ## Tests

  - `pytest test/test_liveness_primer_workflow.py test/test_github_actions_security.py test/test_cicd_*.py`
  - Zizmor reported no findings

### DIFF
--- a/.github/workflows/liveness-primer.yml
+++ b/.github/workflows/liveness-primer.yml
@@ -1,0 +1,85 @@
+name: Analyzer Blast Radius
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - "skylos/**"
+      - "pyproject.toml"
+      - "MANIFEST.in"
+      - ".github/workflows/liveness-primer.yml"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: liveness-primer-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  blast-radius:
+    name: Python dead-code blast radius
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-24.04
+    timeout-minutes: 45
+
+    steps:
+      - name: Check out pinned liveness_primer
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          repository: mcdigman/liveness_primer
+          ref: d6f3118a2cfc465426500eab449005fe56845c58 # v0.1.0
+          path: _liveness_primer
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@e4db8464a088ece1b920f60402e813ea4de65b8f # v4
+        with:
+          version: "0.12.5"
+          python-version: "3.13"
+          enable-cache: false
+
+      - name: Compare base with the pull request merge result
+        env:
+          SKYLOS_REPOSITORY: https://github.com/duriantaco/skylos
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          # The base repository owns this synthetic merge commit, including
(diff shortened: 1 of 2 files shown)


duriantaco:
Thanks @mcdigman for building `liveness_primer` and proposing this integration!

PR #734, opened 2026-08-17
fix(quality): break SKY-C401 clone-type ties deterministically

### LINKED ISSUE #733: Non-deterministic message text in SKY-C401
**Describe the bug**
The message text in SKY-C401 uses a `max` at  [skylos/rules/quality/clones.py:504](https://github.com/duriantaco/skylos/blob/780432d231e3f92e7b143deff1d194df7da1b786/skylos/rules/quality/clones.py#L504) which breaks ties by the iteration order of `set(types)`. As a security feature to mitigate hash collision attacks, python salts the hashes with a per-process randomized PYTHONHASHSEED, so the iteration order is non-deterministic. 

The non-deterministic message text will create pure-noise diffs in future `liveness_primer` CI runs, so the simple fix is to break ties by value/force a canonical sort order, like:
```
clone_type = min(types, key=lambda t: (-types.count(t), t.value)) if types else CloneType.TYPE3
```
instead of 
```
clone_type = max(set(types), key=types.count) if types else CloneType.TYPE3
```
(switched to min of negative count so `type1` beats `type2` as the default message in tie-breaking).

**Note:** with the current pinned `liveness_primer` workflow, the `--quality` checks are not run on any corpus source, so this won't affect current runs, but if/when `--quality` is added it will hit on at least `httpx`.

**To Reproduce**
Reproducer attached distilled from the existing `liveness_primer`/skylos corpus source `httpx/_api.py:L174`.

[skylos_c401_clone_type_hash_order.zip](https://github.com/user-attachments/files/31148203/skylos_c401_clone_type_hash_order.zip)

or run e.g. `skylos --quality fixture/api.py  --select=SKY-C401 --format pretty` repeatedly on:

```
DEFAULT_TIMEOUT_CONFIG = 5.0


def request(method, url, **kwargs):
    """Stand in for the real transport call."""
    return (method, url, kwargs)


def get(
    url,
    *,
    params=None,
    headers=None,
    cookies=None,
    auth=None,
    proxy=None,
    follow_redirects=False,
    timeout=DEFAULT_TIMEOUT_CONFIG,
    verify=True,
    trust_env=True,
):
    """Send a GET request."""
    return request(
        "GET",
        url,
        params=params,
        headers=headers,
        cookies=cookies,
        auth=auth,
        proxy=proxy,
        follow_redirects=follow_redirects,
        timeout=timeout,
        verify=verify,
        trust_env=trust_env,
    )


def options(
    url,
    *,
    params=None,
    headers=None,
    cookies=None,
    auth=None,
    proxy=None,
    follow_redirects=False,
    timeout=DEFAULT_TIMEOUT_CONFIG,
    verify=True,
    trust_env=True,
):
    """Send an OPTIONS request."""
    return request(
        "OPTIONS",
        url,
        params=params,
        headers=headers,
        cookies=cookies,
        auth=auth,
        proxy=proxy,
        follow_redirects=follow_redirects,
        timeout=timeout,
        verify=verify,
        trust_env=trust_env,
    )


def head(
    url,
    *,
    params=None,
    headers=None,
    cookies=None,
    auth=None,
    proxy=None,
    follow_redirects=False,
    timeout=DEFAULT_TIMEOUT_CONFIG,
    verify=True,
    trust_env=True,
):
    """Send a HEAD request."""
    return request(
        "HEAD",
        url,
        params=params,
        headers=headers,
        cookies=cookies,
        auth=auth,
        proxy=proxy,
        follow_redirects=follow_redirects,
        timeout=timeout,
        verify=verify,
        trust_env=trust_env,
    )


def delete(
    url,
    *,
    params=None,
    headers=None,
    cookies=None,
    auth=None,
    proxy=None,
    timeout=DEFAULT_TIMEOUT_CONFIG,
    follow_redirects=False,
    verify=True,
    trust_env=True,
):
    """Send a DELETE request."""
    return request(
        "DELETE",
        url,
        params=params,
        headers=headers,
        cookies=cookies,
        auth=auth,
        proxy=proxy,
        follow_redirects=follow_redirects,
        timeout=timeout,
        verify=verify,
        trust_env=trust_env,
    )
```

and get alternately 
```  api.py · 1 issue

    █  MEDIUM  SKY-C401  Clone group detected (type2, sim=0.995, members=4) examples: api.py:17-43 (function get), api.py:46-72 (function options), api.py:75-101 ...
      Quality  api.py:17
      def get(
```
and (type2 -> type1)
```
  api.py · 1 issue

    █  MEDIUM  SKY-C401  Clone group detected (type1, sim=0.995, members=4) examples: api.py:17-43 (function get), api.py:46-72 (function options), api.py:75-101 ...
      Quality  api.py:17
      def get(
```

**Expected behavior**
Deterministic sort order

**Actual behavior**
Non-deterministic sort order

**Environment**
- Python version: 3.14.6

mcdigman:
Closes #733 

## What does this PR do?

Makes the SKY-C401 clone-group **type label** deterministic. One line in
`_group_connected`; the rest is a regression test.

**AI Use Disclosure**
The issue was isolated by `liveness_primer` running on the corpus source httpx. Claude Opus 5.0 drafted the one-line fix, and it was reviewed by me before pushing. A GPT-5.6 Sol reviewer using repository llm agent guidelines had no findings.

## Why?

`similarity` is the mean over all pairs, so it is unaffected by the label — which
is why this presented as message-only. 

It is *not* purely cosmetic: the finding's
`value` field (`"type1 1.00"` vs `"type2 1.00"`) flips in every case, and a
TYPE2/TYPE3 tie would also flip **severity**, since `analyzer.py` only buckets
TYPE1/TYPE2 into `MEDIUM`.

### Why the httpx group ties

`get`, `options`, `head` and `delete` form one connected component, so six pairs vote:

| pair | text sim | AST sim | type |
| --- | --- | --- | --- |
| `get` / `options` | 0.99194 | 1.00000 | type1 |
| `get` / `head` | 0.99434 | 1.00000 | type1 |
| `options` / `head` | 0.99114 | 1.00000 | type1 |
| `get` / `delete` | 0.91525 | 1.00000 | type2 |
| `options` / `delete` | 0.91231 | 1.00000 | type2 |
| `head` / `delete` | 0.91452 | 1.00000 | type2 |

Exactly 3 type1 and 3 type2 — a perfect tie.

The split arises because the two metrics see different text. `classify_clone`
tries type1 first on `text_norm`, tokenized from the fragment's raw source lines
and therefore including the **signature**; type2 runs on an identifier- and
literal-normalized AST dump of the **body** only. All four bodies are a single
`request(...)` call that normalizes identically, giving AST similarity 1.0
everywhere. In the signatures, `delete` declares `timeout` in a different
position than the other three, dropping its raw token similarity below the 0.98
type1 threshold while the other three stay above it.

## How to test

The new `test/test_clone_type_determinism.py` adds three tests around a fixture
reduced from the httpx group (four request helpers, six pair votes split 3/3):

- `test_fixture_splits_pair_votes_evenly` — guards that the fixture actually
  ties, so the other two tests can't silently stop testing anything.
- `test_tied_vote_prefers_the_stricter_clone_type` — the tie resolves to TYPE1.
- `test_tied_vote_is_stable_across_hash_seeds` — runs the pipeline in a fresh
  interpreter under `PYTHONHASHSEED` 0–7 and asserts all eight agree.

Against the current code, the third test fails on **every** run (seeds 0–7
produce a mix), and the second fails on roughly half of runs. Both pass
consistently with the fix.

```console
$ python -m pytest test/test_clone_type_determinism.py -q
3 passed
```

Both tests pin `_fast_similarity = None`, matching the existing convention in
`test_quality_rules.py`, so results don't depend on whether the Rust extension
is built.

## Checklist

- [x] Tests pass (`python3 -m pytest test/`)
- [x] No new false positives introduced (if modifying analysis logic)

### DIFF
--- a/skylos/rules/quality/clones.py
+++ b/skylos/rules/quality/clones.py
@@ -501,7 +501,11 @@ def _group_connected(
                     types.append(p.clone_type)
 
         avg_sim = sum(sims) / len(sims) if sims else cfg.grouping_threshold
-        clone_type = max(set(types), key=types.count) if types else CloneType.TYPE3
+        clone_type = (
+            min(set(types), key=lambda t: (-types.count(t), t.value))
+            if types
+            else CloneType.TYPE3
+        )
 
         groups.append(
             CloneGroup(
--- a/test/test_clone_type_determinism.py
+++ b/test/test_clone_type_determinism.py
@@ -1,0 +1,254 @@
+"""SKY-C401 clone-type labels must not depend on the interpreter hash seed.
+
+`_group_connected` picks a group's clone type by majority vote over its pair
+types. When that vote ties, the winner used to be decided by `set` iteration
+order, which CPython randomizes per process via `PYTHONHASHSEED`, so identical
+input could be reported as `type1` on one run and `type2` on the next.
+
+The fixture below is reduced from `httpx/_api.py`, where the tie was first
+observed: four request helpers whose bodies normalize identically (so every
+pair is type2-similar), three of which share a keyword-only parameter order
+(so those three pairs are also type1-similar). That splits the six pair votes
+exactly 3/3.
+"""
+
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+import pytest
+
+from skylos.rules.quality import clones as clones_mod
+
+HASH_SEEDS = ("0", "1", "2", "3", "4", "5", "6", "7")
+
+CLONE_SOURCE = '''\
+DEFAULT_TIMEOUT_CONFIG = 5.0
+
+
+def request(method, url, **kwargs):
+    return (method, url, kwargs)
+
+
+def get(
+    url,
+    *,
+    params=None,
+    headers=None,
+    cookies=None,
+    auth=None,
+    proxy=None,
+    follow_redirects=False,
+    timeout=DEFAULT_TIMEOUT_CONFIG,
+    verify=True,
+    trust_env=True,
+):
+    """Send a GET request."""
+    return request(
+        "GET",
+        url,
+        params=params,
+        headers=headers,
+        cookies=cookies,
+        auth=auth,
+        proxy=proxy,
+        follow_redirects=follow_redirects,
+        timeout=timeout,
+        verify=verify,
+        trust_env=trust_env,
+    )
+
+
+def options(
+    url,
+    *,
+    params=None,
+    headers=None,
+    cookies=None,
+    auth=None,
+    proxy=None,
+    follow_redirects=False,
+    timeout=DEFAULT_TIMEOUT_CONFIG,
+    verify=True,
+    trust_env=True,
+):
+    """Send an OPTIONS request."""
+    return request(
+        "OPTIONS",
+        url,
+        params=params,
+        headers=headers,
+        cookies=cookies,
+        auth=auth,
+        proxy=proxy,
+        follow_redirects=follow_redirects,
+        timeout=timeout,
+        verify=verify,
+        trust_env=trust_env,
+    )
+
+
+def head(
+    url,
+    *,
+    params=None,
+    headers=None,
+    cookies=None,
+    auth=None,
+    proxy=None,
+    follow_redirects=False,
+    timeout=DEFAULT_TIMEOUT_CONFIG,
+    verify=True,
+    trust_env=True,
+):
+    """Send a HEAD request."""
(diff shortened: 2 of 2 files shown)


duriantaco:
close PR #734 as superseded by merged PR #756 (https://github.com/duriantaco/skylos/pull/756)

PR #736, opened 2026-08-18
fix(ai-defect): avoid oversized registry metadata false negatives

### LINKED ISSUE #735: SKY-D225 drops findings for large npm responses
**Describe the bug**
When skylos calls npm, if a package exists but npm response is >1 MB, easily possibly/common for dozens of large common libraries, including e.g. `react`, skylos just silently drops analyzing that package for SKY-D225 and some other similar rules (including caching of the results of the commands), rather than failing closed with an error or any other handling. For some findings this *can* appear to produce the correct behavior accidentally, for others not.

This could be mitigated by trying to get a file that is expected to be smaller, e.g at [manifest_dependency_hallucination.py (line 1992)](https://github.com/duriantaco/skylos/blob/780432d231e3f92e7b143deff1d194df7da1b786/skylos/rules/ai_defect/manifest_dependency_hallucination.py#L1992), building a command that will become:

```GET https://registry.npmjs.org/{package}/{version}```

instead of 

```GET https://registry.npmjs.org/{package}```.

Or, better, just use HEAD, since the response *body* is actually not needed at all if the top version is used instead.

**perf effects of a fix:**
1. Fetching files that are expected to be smaller means that they can be written to the cache (whereas large ones cannot), improving warm-cache performance of skylos. Cache-misses for commonly pinned repos like `react` due to the 1 Mb cutoff are currently the single largest waste of time *in the entire skylos analysis* when running skylos with --ai-defects enabled.
2. Fetching files that are expected to be << 1 MB reduces disk space wasted by the cache.

**Other similar issues to harden in the same PR**

The same 1 MB limit applies to pypi and Go as well. Several common/major pypi packages actually have json size>1 MB and so the defect will also hit there, including numpy, scipy, and pandas. I'm not sure any common Go packages have such a large response, but the hardening may as well land there as well.

**To Reproduce**

`package.json`:

```
{
  "dependencies": {
    "through2": "999999.0.0"
  }
}
```

vs 
```
{
  "dependencies": {
    "react": "999999.0.0"
  }
}
```

neither exists, but only 
```
  control/package.json · 1 issue

    █  HIGH  SKY-D225  Hallucinated npm dependency version 'through2@999999.0.0'. Version does not exist in the npm registry.
      AI Defect  control/package.json:3
      "through2": "999999.0.0"
```
is emitted.

[skylos_d225_large_npm_response.zip](https://github.com/user-attachments/files/31153934/skylos_d225_large_npm_response.zip)

**Expected behavior**
SKY-D225 in both cases.

**Actual behavior**
SKY-D225 only when the npm response is < 1 Mb.

**Environment**
- Python version: 3.14.6

### LINKED ISSUE #737: FP: SKY-D225 fires on versions that are actually resolvable
**What was flagged?**
- File: package.json
- Category: [ai-defects]

**Context**
Semi-latent false positive surfaced/made much more common due to fixes in #736 causing SKY-D225 to not simply drop entire common dependencies.

SKY-D225 strips e.g. `>=` from version requirements, so it fires on cases like `through2>=2`, while through2=5.0.11 is current so that requirement *is* resolvable.

**Skylos version**
skylos 4.33.2

**Minimal reproduction**
```
{
  "dependencies": {
    "through2": ">=2",
    "react": ">=16"
  }
```

gives (without #736)
```
  reproduce/package.json · 1 issue

    █  HIGH  SKY-D225  Hallucinated npm dependency version 'through2@2'. Version does not exist in the npm registry.
      AI Defect  reproduce/package.json:3
      "through2": ">=2",
```

and with #736 it would also fire on `react`, which would be a bad false positive to introduce because it is very common as a dependency; therefore, #736 should not be merged until a fix for this issue is available as well.

[skylos_d225_shorthand_range_base.zip](https://github.com/user-attachments/files/31161361/skylos_d225_shorthand_range_base.zip)

### PR DESCRIPTION
**Pull request by mcdigman**
The first fix surfaced a latent false-positive that would cause noise downstream; they need to be fixed in the same release version to avoid that, so I thought to just combine them into the same PR to guarantee that.

Closes #735
Closes #737 

## Second issue surfaced
#737 was quasi-latent before the original fix, but would have gotten worse due to it becoming more apparent on specifically large packages; for example, in addition to the #737 reproducer, the original fix would have cause a FP in skylos itself at `editors/vscode/package.json`:
```
+  SKY-D225  NA  HIGH      ai_defect  editors/vscode/package.json:L16  typescript@5.0.0  Hallucinated npm dependency version 'typescript@5.0.0'. Version   -
                                                                                         does not exist in the npm registry.
  ```
  which is caused because `"typescript": "^5.0.0"` specified there *actually* does not exist exactly (5.0.0 was a dev version and has a corresponding qualifier in the version name), but npm still resolves it fine anyway, to 5.9.x.
  
 The fix approach I chose here is the one that minimizes false positives, which is to only have SKY-D225 fire on npm for exact pins; other rules can fire on ranges. Note that part of the fix isn't really needed for pypi, because the pypi api uses PEP 440 normalization which will respond as expected in most stripped cases unless it is given a reference version that actually does not exist, in which case firing SKY-D225 seems fine.
 
 **Use of HEAD**
 Using HEAD instead of GET eliminates the 1 MB fix completely, and maybe closes some incidental network edge cases; it is also probably better/non-inferior performance-wise, although I haven't tested the profiling impact as this isn't specifically a perf PR.
 
**AI Use Disclaimer**

Codex GPT-5.6 Sol drafted this fix, and it was reviewed according to skylos repository standards by Claude Fable 5.0 and a separate GPT-5.6 Sol Agent. I inspected the code fixes manually and they look as expected.
## Summary

- query version-scoped registry resources instead of full npm package metadata
- use bodyless `HEAD` probes for npm, PyPI, and Go version/package existence checks
- distinguish npm SemVer ranges from exact pins, checking ranges for package existence only
- preserve authoritative present/missing cache entries without caching unknown results
- remove the now-unused registry response body reader and its 1 MB cutoff

## Reproduction

Before this change, scanning `numpy==999999.0.0` returned no AI defect because the version request returned 404 and the 3,616,753-byte package-wide PyPI response exceeded the 1 MB limit. That degraded the result to `unknown`, suppressed `SKY-D225`, and did not warm the cache.

After this change, the same real-registry CLI reproduction reports `SKY-D225` and records `PyPI:numpy:999999.0.0 = missing_version`.

Live probes also confirmed the expected `HEAD` 200/404 behavior for representative existing and missing npm versions, PyPI versions/packages, and Go `.info`/`@v/list` resources.

Issue #737 also reproduced on the PR head: `through2: ">=2"` and `react: ">=16"` were reduced to literal versions `2` and `16`, producing false `SKY-D225` findings. Range specs are now preserved and use package-only probes. Exact full SemVer pins still receive version-level validation, while missing ranged packages and suspicious existing lookalikes remain detectable. A real scan of `editors/vscode/package.json` likewise produces no finding for `typescript: "^5.0.0"` and caches that range as a present package.

## Tests

- focused manifest/parser/diff suites — 60 passed
- full suite — 7,071 passed, 7 skipped; two unrelated existing agent-behavior error-message assertions failed

Regression coverage verifies:

- oversized npm version metadata cannot suppress suspicious-existing lookalike detection
- npm, PyPI, and Go missing-version findings survive to scanner output
- package-wide fallback bodies are never read
- authoritative results are persisted and reused without a second registry request
- npm ranges do not produce literal-version `SKY-D225` false positives
- ranged missing packages and suspicious lookalikes still produce `SKY-D222`

No no-network behavior is changed by this PR.

### DIFF
--- a/skylos/rules/ai_defect/manifest_dependency_hallucination.py
+++ b/skylos/rules/ai_defect/manifest_dependency_hallucination.py
@@ -1989,22 +1989,31 @@ def _check_npm_version(name: str, version: str) -> str:
     if package_path is None:
         return STATUS_UNKNOWN
 
-    url = f"{NPM_REGISTRY_ORIGIN}/{package_path}"
+    raw_version = version.strip()
+    if not raw_version:
+        return STATUS_UNKNOWN
+
+    safe_version = quote(raw_version, safe="")
+    version_url = f"{NPM_REGISTRY_ORIGIN}/{package_path}/{safe_version}"
     try:
-        data = _fetch_json(url, user_agent="skylos-npm-dep-scanner/1.0")
+        _fetch_json(version_url, user_agent="skylos-npm-dep-scanner/1.0")
+        return STATUS_PRESENT
+    except urllib.error.HTTPError as exc:
+        if exc.code != 404:
+            return STATUS_UNKNOWN
+    except (urllib.error.URLError, TimeoutError, OSError, ValueError):
+        return STATUS_UNKNOWN
+
+    package_url = f"{NPM_REGISTRY_ORIGIN}/{package_path}"
+    try:
+        _fetch_head(package_url, user_agent="skylos-npm-dep-scanner/1.0")
+        return STATUS_MISSING_VERSION
     except urllib.error.HTTPError as exc:
         if exc.code == 404:
             return STATUS_MISSING_PACKAGE
         return STATUS_UNKNOWN
     except (urllib.error.URLError, TimeoutError, OSError, ValueError):
         return STATUS_UNKNOWN
-
-    versions = data.get("versions")
-    if not isinstance(versions, dict):
-        return STATUS_UNKNOWN
-    if version in versions:
-        return STATUS_PRESENT
-    return STATUS_MISSING_VERSION
 
 
 def _check_go_version(name: str, version: str) -> str:
@@ -2138,6 +2147,19 @@ def _fetch_json(url: str, *, user_agent: str) -> dict[str, Any]:
     if isinstance(data, dict):
         return data
     return {}
+
+
+def _fetch_head(url: str, *, user_agent: str) -> None:
+    if not _allowed_registry_url(url):
+        raise ValueError("Registry URL host is not allowed")
+
+    request = urllib.request.Request(url, method="HEAD")
+    request.add_header("User-Agent", user_agent)
+    with urllib.request.urlopen(  # skylos: ignore[SKY-D216] URL is validated against fixed registry hosts above.
+        request,
+        timeout=5,
+    ):
+        pass
 
 
 def _fetch_text(url: str, *, user_agent: str) -> str:
--- a/skylos/rules/ai_defect/manifest_dependency_hallucination.py
+++ b/skylos/rules/ai_defect/manifest_dependency_hallucination.py
@@ -29,6 +29,7 @@ from skylos.rules.sca.vulnerability_scanner import (
     ECOSYSTEM_GO,
     ECOSYSTEM_NPM,
     ECOSYSTEM_PYPI,
+    _extract_exact_npm_version,
     parse_go_mod,
     parse_package_json_candidates,
     parse_pyproject_toml_candidates,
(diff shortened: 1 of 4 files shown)


duriantaco:
superseded by #751 and #752.

PR #744, opened 2026-08-19
perf(dead-code): index documented method references

### LINKED ISSUE #743: Performance: Index method references in documentation scans to avoid re-scans
When scanning large repos, one of the single largest time-wasters according to cProfile is `_docs_reference_method` running four full-text regex searches for each method; similar to the batching performance improvements recently realized for grep verification, we can index the results of these scans so they only need to be done once.

On potential corpus source xarray, such batching improves overall performance of the entire scan by ~40%

**Alternatives considered**
Could use rg like the grep verification.

mcdigman:
Closes #743

**AI Use Disclosure** This fix was based on a cProfile run of skylos on xarray. Codex GPT-5.6 Sol drafted the fix and it was reviewed according to repository review guidelines by Claude Fable 5 and a separate GPT-5.6 Sol agent. `liveness_primer` shows 0 diffs on an expanded test corpus beyond what is currently used in the CI. I reviewed the code fixes myself.

## Summary

- index qualified method references in documentation in one pass instead of rescanning the complete documentation corpus for every candidate
- preserve the existing qualified-name, Sphinx `:meth:` suffix, long underscore-call, Unicode identifier, and overlapping escaped-name behavior
- keep independent legacy-equivalent checks for rare identifiers outside Python regex `\w` semantics
- add direct Unicode-prefix and end-to-end Kotlin backtick regressions

## Measurement

Base: `a5d0319499b8429ebfa980f82626d7426a24dece`
Corpus: xarray `c2998a75ef32f6cf4d62cd946f2200ca953550d9` (`SKYLOS_JOBS=1`, `--no-grep-verify`)

- documentation rescue helper: 9.3943s -> 0.0275s (342x)
- latest complete scan wall time: 29.36s -> 20.76s (1.41x)
- all six dead-code finding buckets matched exactly
- all 680 liveness rescues matched exactly

## Tests

- `python -m pytest -q -p no:cacheprovider test/test_dead_code_liveness.py` (12 passed)
- `python -m pytest -q -p no:cacheprovider test/test_dead_code*.py test/test_framework_aware.py` (83 passed)
- self-scan confirms the new helpers add no SKY-Q301/SKY-Q306 findings

### DIFF
--- a/skylos/deadcode/liveness.py
+++ b/skylos/deadcode/liveness.py
@@ -496,10 +496,9 @@ def _qualified_documented_method_keys(
         for name, keys in qualified.items()
         if keys - word_candidates
     }
-    if unusual_qualified:
-        qualified_pattern = _literal_alternation(unusual_qualified)
-        for match in re.finditer(rf"(?=\b({qualified_pattern})\b)", text):
-            referenced.update(unusual_qualified[match.group(1)])
+    for qualified_name, keys in unusual_qualified.items():
+        if re.search(rf"\b{re.escape(qualified_name)}\b", text):
+            referenced.update(keys)
     return referenced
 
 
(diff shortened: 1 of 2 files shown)


duriantaco (approved):
Found a legacy-equivalence gap in the new Sphinx :meth: indexing. A malformed role could consume a later valid role, causing a documented method to be reported as unused.

I pushed ba3b202, which uses overlapping role detection. The rest LGTM. Thanks!

PR #746, opened 2026-08-19
perf(ai-defect): batch API surface cache I/O

### LINKED ISSUE #745: Perf:  Validate/write API surface cache once per scan
When scanning large repositories like xarray, numpy, or scipy, significant time is wasted currently wasted at e.g. `cache_python_api_surface()` (line 85) and `cache_api_symbol_surface()` (line 97) repeatedly loading, normalizing and rewriting the API surface cache multiple times per scan. After #743, fixing this is one of the single biggest/easiest perf wins identified in a cProfile scan.

**Proposed solution**
It is a relatively small behavior-preserving fix to load the cache once, collect missing surfaces, and update it only once per scan. This will improve cold-cache performance significantly, and 

**Alternatives considered**
Leaving the slower behavior

**Note for future reference** 
For the biggest large repositories I have tested like scipy and numpy, the API surface cache can be >3 MiB approaching the 10 MiB limit in size, so alternate handling or a cache size lift may be necessary to handle that in the future. This could be a separate future issue.

mcdigman:
Closes #745.

**AI Use Disclaimer**
This fix was drafted by Codex GPT-5.6 Sol and reviewed per skylos repository review guidelines by Claude Fable 5 and a separate GPT-5.6 Sol agent. I reviewed the code changes. a `liveness_primer` run with an expanded corpus and including ai-defects shows 0 diffs. 

## Summary

- load the shared API truth cache once per API-signature scan instead of once per imported module
- accumulate newly captured Python API surfaces in memory and flush each cache file once at the end of the scan
- detect cache files changed by another writer, refresh the shared lookup snapshot, and merge pending entries without overwriting newer same-key values
- skip AST parsing when source text cannot contain an allowlisted import root, with NFKC normalization for non-ASCII Python identifiers
- retain lazy initialization, cache validation, environment-key behavior, and the existing on-exception flush behavior

## Correctness follow-up

At `34c904300aa5ff3b3a03b4a780eca01042955912`, an update made after the batch snapshot could be missed during verification and then overwritten in both `python_api_surface.json` and `api_symbol_truth.json`. The regression test reproduces the resulting extra `pkg_b.known` finding and lost `pkg_b` cache entries. The follow-up refreshes a changed shared snapshot and rebases pending additions onto the current files at flush time. It now reports only `pkg_a.missing` and retains `pkg_b` in both caches.

The source prefilter also covers Python identifier normalization: `import 𝕡andas as pd` is parsed by Python as an import of `pandas`, so non-ASCII source is NFKC-normalized before a file is skipped.

## Measurement

Base: `a5d0319499b8429ebfa980f82626d7426a24dece`
Head: `022ddd769aaf7eff1fbd220defbf1ec5200eb238`
Corpus: SciPy `0e0ea9350596693dcc3e3604ef8c6dde220c8b2c`, 994 Python files

`scan_python_api_signature_hallucinations` is called directly in a fresh process per run, isolating the rule from an unrelated nonzero exit in full-CLI SciPy scans. Cold runs wipe `.skylos` first; three interleaved trials per cell; medians reported.

| Config | base cold | head cold | base warm | head warm |
|---|---|---|---|---|
| default allowlist | 4.47s | 0.71s (6.3x) | 3.69s | 0.34s (10.7x) |
| `numpy` allowlist (18 modules; 816/994 files mention `numpy`) | 10.09s | 4.30s (2.3x) | 5.29s | 3.65s (1.4x) |

The `numpy` config isolates the cache-batching win (the prefilter skips few files there); the default config isolates the prefilter win (SciPy imports pandas in one file, requests in none, so the base cost is almost entirely parse/visit of irrelevant files).

- finding-identity sets are identical across all benchmark runs, including earlier trials at `34c9043` (24 runs under the `numpy` config with 4 SKY-D224 findings each; 16 under the default with 0)
- cold-run cache files are identical between variants: same 18 module keys, content equal except `captured_at` timestamps and three `repr()`-embedded default-value memory addresses that vary between any two runs of either variant
- a differential fuzz of the parse prefilter over 17 unicode-tricky sources (fullwidth and mathematical-alphanumeric identifiers, combining-mark suffixes, string/comment-only mentions, non-ASCII docstrings) produced identical findings on base and head
- warm profile at head under the `numpy` config: the entire cache path is a single load+normalize, roughly 3.5% of scan time; the remainder is `ast.parse` plus AST visiting
- the cold-plus-warm cache I/O regression test observes one Python-cache load, two shared-cache loads, and one save per cache

## Tests

- `python -m pytest -q -p no:cacheprovider test/test_api_signature_hallucination.py test/test_python_api_surface.py test/test_api_symbol_truth.py` (23 passed)
- Ruff import checks and `git diff --check` pass
- Pyright reports only the same existing `importlib.util` diagnostic present at the base
- both CI test matrices complete with only the pre-existing simulation-branch failure in `test_get_git_changed_files_unions_worktree_and_branch_changes`; the prior `34c9043` run fails on the identical test
- title, analyzer speed, both Docker smokes, corpus guard, advisory quality, quality benchmark, and the Skylos advisory scan pass

### DIFF
--- a/skylos/core/api_symbol_truth.py
+++ b/skylos/core/api_symbol_truth.py
@@ -3,11 +3,15 @@ from __future__ import annotations
 from pathlib import Path
 from typing import Any
 
-from skylos.core.safe_cache_io import load_project_json_cache, save_project_json_cache
-
+from skylos.core.safe_cache_io import (
+    load_project_json_cache,
+    project_cache_lock,
+    save_project_json_cache,
+)
 
 API_SYMBOL_TRUTH_SCHEMA_VERSION = 1
 API_SYMBOL_TRUTH_CACHE_PATH = Path(".skylos") / "cache" / "api_symbol_truth.json"
+API_SURFACE_CACHE_LOCK_PATH = Path(".skylos") / "cache" / ".api_surface.lock"
 MAX_API_SYMBOL_TRUTH_BYTES = 10_000_000
 
 SURFACE_KIND_PYTHON_MODULE = "python_module"
@@ -62,6 +66,29 @@ def save_api_symbol_truth_cache(
     *,
     cache_path: str | Path = API_SYMBOL_TRUTH_CACHE_PATH,
 ) -> bool:
+    """Replace the full cache. Use cache_api_symbol_surface for merged updates."""
+
+    if not _valid_cache_payload(payload):
+        return False
+    with project_cache_lock(
+        project_root,
+        API_SURFACE_CACHE_LOCK_PATH,
+    ) as lock_acquired:
+        if not lock_acquired:
+            return False
+        return _save_api_symbol_truth_cache_unlocked(
+            project_root,
+            payload,
+            cache_path=cache_path,
+        )
+
+
+def _save_api_symbol_truth_cache_unlocked(
+    project_root: str | Path,
+    payload: dict[str, Any],
+    *,
+    cache_path: str | Path = API_SYMBOL_TRUTH_CACHE_PATH,
+) -> bool:
     if not _valid_cache_payload(payload):
         return False
     payload = {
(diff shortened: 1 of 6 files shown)


duriantaco:
Thanks for the PR. The batching approach works and fixes the repeated cache I/O.

 I pushed a follow-up commit (f2e66ff) to fix two edge cases, them being:
1. concurrent cache writers could overwrite each other
2. very large custom module lists could make the prefilter much slower
  
Cache should now lock, reload, merge, and save safely. Large module lists skip only the prefilter.

duriantaco (approved):

PR #747, opened 2026-08-19
fix(analyzer): bound oversized fixed-string grep batches

### LINKED ISSUE #742: Skylos grep verification fails on large repositories
**Describe the bug**
On a large repository, a grep verification result can honestly exceed the current cap of 8 MiB. For example, a grep verification pass verifying the symbol `numpy` in `numpy/numpy` returns a 10,452,792 byte JSON, causing skylos to exit with an error when asked to scan `numpy/numpy`.

Failing on scanning large popular repos is undesirable. This would also affects other larger repos like scipy.

While one solution could be to raise the 8 MiB cap a bit or scale it to the actual size of the repository, a potentially more robust fix is to just short-circuit grep verification when vastly more than adequate evidence of its usage is accumulated (as occurs when scanning the repo `numpy` for the string `numpy`). 

I will PR a demo of the short-circuit fix shortly. 

**To Reproduce**
Run skylos on numpy/numpy.

**Expected behavior**
Exit with no error code.

**Actual behavior**
Error code 2.

**Environment**
- Python version: 3.14.6

mcdigman:
Closes #742 

**AI Use Disclosure**
The initial commits were generated by Codex GPT-5.6 Sol and some follow-ups were done by Claude. Review was done according to skylos repository guidelines using Claude Fable 5 and a separate Codex GPT-5.6 Sol agent. I reviewed the changes myself.  I verified skylos can now run without failure on numpy and scipy. `liveness_primer` with an expanded corpus shows 0 diffs.

## What changed
- Retry fixed-string batches that exceed the aggregate output cap using a bounded, streamed search.
- Inventory and Python-sort eligible files, then search bounded explicit-path chunks with one ripgrep worker.
- Retain only each request canonical 256-result evidence window and stop scheduling chunks once every request is saturated.
- Rebuild each chunk's pattern file from the still-active requests only, so a saturated request's pattern is no longer searched, and partition mixed chunks so one direct-grep pattern does not veto the ordered retry for the rest of its chunk.
- Account a retained match once, as the bytes of the evidence string it contributes to results, against a dedicated retained-data constant. The earlier accounting summed the match's path, content, and evidence, roughly twice the evidence bytes, so a full batch of hot patterns on a scipy-sized tree sat within 12% of the cap and crossed it under a longer checkout prefix, discarding a nearly finished sweep and redoing all of it in split halves.
- Hold a file's matches until ripgrep reports its end event and refuse the sweep when that event carries a binary offset. Ripgrep quits at the first NUL byte in a file it walked to but converts NUL to a line terminator in a path named explicitly, so an explicit-path sweep would otherwise return matches the canonical batch never produces, at line numbers shifted by the conversion.
- Resolve the ordered inventory at most once per batch, since the split cascade retries the ordered path at every node and a failing inventory otherwise costs 2N-1 full-tree walks.
- Fail closed to the existing split path if inventory, ordering, binary, retention, process, or deadline checks cannot complete.

## Regression coverage
- Forces the original aggregate output-cap path.
- Proves exact canonical output across the same-prefix file/directory ordering edge case.
- Covers successful ordered retry and conservative fallback when inventory is unavailable.
- Pins single-counted retention accounting, and abstention when one file or one sweep exceeds the retained-data limit.
- Covers the binary-conversion refusal, including the case where a request saturates on the converted file, plus refusal of a truncated stream and of a malformed file end event.
- Runs real ripgrep down both the canonical and ordered paths over one tree and asserts the results are identical.
- Covers the file inventory directly, including its exclusion globs, single-file roots, and the sort-key collision guard.
- Asserts the inventory is resolved once for a batch that splits, and that a deadline failure is not cached as a permanent one.

## Validation
- test/test_grep_verify.py: 173 passed
- Full suite: 7085 passed, no new failures
- Pyright: 0 errors; ruff check clean
- Differential against the canonical batch with the ordered chunk size forced to 3: 600 randomized trees without NUL bytes agree exactly with no abstentions; 400 trees seeded with NUL bytes agree exactly, abstaining 175 times
- pydantic, 128 hottest identifiers, output cap lowered to force the retry: the ordered retry answers 128/128 requests in 0.38s where splitting answers 24/128 in 2.0s
- NumPy commit 6d743ffd3adee7cc754fd0267948ff60ac6bdc23 with SKYLOS_JOBS=1: exit 0, parseable 91.6 MB JSON, analysis_error_count 0

Base SHA: 780432d
Head SHA: db88898

### DIFF
--- a/skylos/core/grep_verify_common.py
+++ b/skylos/core/grep_verify_common.py
@@ -89,6 +89,12 @@ _GREP_BATCH_MAX_STDERR_BYTES = 128 * 1024
 _GREP_BATCH_MAX_MATCHES = _GREP_BATCH_SIZE * 256
 _GREP_ORDERED_MAX_PATHS = 64
 _GREP_ORDERED_MAX_PATH_BYTES = 64 * 1024
+# Bounds the canonical evidence an ordered sweep may retain, accounted as
+# the evidence bytes each distinct match contributes to results. Counting
+# a match's fields beyond its evidence would halve the usable window and
+# push full hot batches on scipy-sized trees into a pointless sweep that
+# is then redone in split halves.
+_GREP_ORDERED_MAX_RETAINED_BYTES = _GREP_BATCH_MAX_OUTPUT_BYTES
 _GREP_UNICODE_MAX_INPUT_BYTES = 512 * 1024
 # Every request in a bounded batch must receive the same Unicode-boundary
 # adjudication. Skipping later requests makes a successful batch partial and
@@ -976,12 +982,8 @@ def _record_ordered_fixed_string_match(
     ]
     if not matched_requests:
         return
-    retained_bytes = (
-        len(match.path.encode("utf-8"))
-        + len(match.content.encode("utf-8"))
-        + len(match.evidence.encode("utf-8"))
-    )
-    if search.retained_bytes + retained_bytes > _GREP_BATCH_MAX_OUTPUT_BYTES:
+    retained_bytes = len(match.evidence.encode("utf-8"))
+    if search.retained_bytes + retained_bytes > _GREP_ORDERED_MAX_RETAINED_BYTES:
         raise _GrepOutputLimitExceeded(
             "ordered ripgrep matches exceeded the retained-data limit"
         )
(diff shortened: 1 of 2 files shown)


duriantaco:
#757 closes the issue. Thanks for your input!

PR #748, opened 2026-08-19
fix(sca): scope npm dependency line anchoring to its declaring section

### LINKED ISSUE #738: SKY-D225 Incorrect Line assignment
**Context**
Found while creating #736, but unrelated fix/preexisting and would not block merge of my existing PR; will let the maintainer assess the scope of changing finding line identity/the other fixes instead of proposing my own fix to this one for now.

**Describe the bug**
SKY-D225 can assign the incorrect line number to a finding, resulting in an unreadable diff. 

**To Reproduce**
```
{
  "name": "line-anchor-mre",
  "keywords": [
    "through2"
  ],
  "dependencies": {
    "through2": "999999.0.0"
  }
}
```
gives:
```
  package.json · 1 issue

    █  HIGH  SKY-D225  Hallucinated npm dependency version 'through2@999999.0.0'. Version does not exist in the npm registry.
      AI Defect  package.json:4
      "through2"
```

When, actually, line 7 should have been flagged instead of line 4. This is apparently due to _find_line in skylos/rules/sca/vulnerability_scanner.py matching the first occurrence in the file. This makes the code snippets much less usable, especially if the relevant lines are separated by a larger amount.

[skylos_d225_first_occurrence_line_anchor.zip](https://github.com/user-attachments/files/31162623/skylos_d225_first_occurrence_line_anchor.zip)

**Expected behavior**
Match line 7.

**Actual behavior**
Matches line 4.

**Environment**
- Python version: 3.14.6

AmirF194:
## What does this PR do?

`_find_line()` anchors an SCA finding to the first textual occurrence of the
needle string anywhere in the manifest file, not the declaring entry. The npm
path in `_parse_package_json` calls it unscoped (`_find_line(txt, f'"{name}"')`),
so any earlier occurrence of the same quoted string elsewhere in the file
(a `keywords` entry, a description, the same name in the other dependency
section) hijacks the reported line. `_poetry_dependency_source` already avoids
this for pyproject.toml with a section-scoped line walk; the npm path had no
equivalent.

Adds `_npm_dependency_source()`, mirroring the poetry helper's shape: track
whether the walk is inside the `dependencies`/`devDependencies` object being
parsed, match the key only within that scope, and fall back to `_find_line`
if no scoped match is found.

## Why?

Fixes #738. SKY-D225 (hallucinated npm dependency version) reported line 4
for the issue's repro when the actual `through2` entry was at line 7, because
line 4 held an unrelated `"through2"` string in a `keywords` array. Wrong line
numbers make the finding's code snippet unusable for diff review.

## How to test

```
python -m pytest test/test_sca_cache.py -k npm_finding_line -v
```

Both new tests fail on `main` and pass on this branch:
- `test_npm_finding_line_anchors_to_dependency_entry_not_first_occurrence`:
  the issue's own repro (name repeated in `keywords`, line 4 vs line 7).
- `test_npm_finding_line_distinguishes_same_name_in_dependencies_and_devdependencies`:
  the same dependency name declared in both sections previously collapsed to
  one line.

## Precision Impact

Output-only: reported `line`/`snippet` for existing findings. No change to
which dependencies are flagged or their version comparison.

## Checklist

- [x] Tests pass (`python3 -m pytest test/`): full suite, both before and
      after this change, 7027 passed / 46 skipped / 3 pre-existing failures
      unrelated to this file (`test_agent_remediate`, `test_dead_code_benchmark`,
      `test_security_benchmark`; identical on `main`, traced to the Go
      analyzer engine not being built in this environment, not to this diff).
- [ ] No new false positives (N/A, line attribution only, not a detection change)
- [ ] Corpus case (N/A, not a precision/detection change)
- [ ] CHANGELOG (auto-generated by release-please from this PR's commits)

Not verified: the repo's Go-engine-dependent CI legs (`docker-smoke`,
`analyzer-speed`, `corpus`, `parity`, `quality-benchmark`) and the full
`test-matrix` install target, since building the Go engine and Rust toolchain
was out of scope for a Python-only SCA parsing fix. `skylos/rules/sca/`
does not touch either engine.

### DIFF
--- a/skylos/rules/sca/vulnerability_scanner.py
+++ b/skylos/rules/sca/vulnerability_scanner.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
 import json
+import logging
 import os
 import re
-import logging
+from json.decoder import scanstring
 from pathlib import Path
 
 try:
@@ -536,25 +537,77 @@ def parse_pyproject_toml_candidates(path: Path) -> list[dict]:
     )
 
 
-def _npm_dependency_source(text: str, section: str, name: str) -> tuple[int, str]:
-    in_section = False
+_NPM_JSON_STRUCTURAL_RE = re.compile(r'["{}\[\]:,]')
+
+
+def _npm_dependency_lines(
+    text: str, sections: tuple[str, ...]
+) -> dict[tuple[str, str], int]:
+    """Map (section, dependency name) to the 1-indexed line of its declaring key.
+
+    Walks the raw text via its own structural characters rather than line-by-line
+    (json.loads exposes no position info), so a section object written on a single
+    line -- open and close brace on the same line as its entries -- is anchored
+    exactly like a multi-line one; the prior line-scoped scan closed such a section
+    immediately after opening it and fell back to the first textual occurrence of
+    the name anywhere in the file. A key is only recorded while directly inside a
+    TOP-LEVEL object literally named by `sections` (depth == section_depth + 1), so
+    a same-named key nested elsewhere (an `overrides` block, a description) cannot
+    hijack the line, and `sections` never includes `overrides` here in the first
+    place -- this file does not scan it.
+    """
+    lines: dict[tuple[str, str], int] = {}
     depth = 0
-    section_key = f'"{section}"'
-    for lineno, raw_line in enumerate(text.splitlines(), 1):
-        line = raw_line.strip()
-        if not in_section:
-            if line.startswith(section_key) and ":" in line:
-                in_section = True
-                depth = line.count("{") - line.count("}")
-                if depth <= 0:
-                    in_section = False
+    section: str | None = None
+    section_depth = 0
+    pending_key: str | None = None
+    expect_value = False
+    scanned = 0
+    lineno = 1
+    pos = 0
+
+    while True:
+        match = _NPM_JSON_STRUCTURAL_RE.search(text, pos)
+        if match is None:
+            return lines
+        char = match.group()
+        pos = match.end()
+
+        if char == '"':
+            key_start = match.start()
+            try:
+                key, pos = scanstring(text, pos)
+            except ValueError:
+                return lines
+            if expect_value:
+                pending_key = None
+                expect_value = False
+                continue
+            pending_key = key
+            if section is not None and depth == section_depth + 1:
+                lineno += text.count("\n", scanned, key_start)
+                scanned = key_start
+                lines[(section, key)] = lineno
             continue
-        if line.startswith(f'"{name}"') and ":" in line:
-            return lineno, line
-        depth += line.count("{") - line.count("}")
-        if depth <= 0:
-            in_section = False
-    return _find_line(text, f'"{name}"'), name
+
+        if char == ":":
+            expect_value = True
+            continue
+        if char == ",":
+            pending_key = None
+            expect_value = False
+            continue
+        if char in "{[":
+            if char == "{" and depth == 1 and expect_value and pending_key in sections:
+                section = pending_key
+                section_depth = depth
+            depth += 1
+        else:
+            depth -= 1
+            if section is not None and depth <= section_depth:
+                section = None
+        pending_key = None
+        expect_value = False
 
 
 def _parse_package_json(path: Path, *, allow_version_ranges: bool) -> list[dict]:
@@ -571,6 +624,8 @@ def _parse_package_json(path: Path, *, allow_version_ranges: bool) -> list[dict]
         data = json.loads(txt)
     except Exception:
         return deps
+
+    dependency_lines = _npm_dependency_lines(txt, ("dependencies", "devDependencies"))
 
     for section in ("dependencies", "devDependencies"):
         pkg_deps = data.get(section, {})
@@ -593,7 +648,9 @@ def _parse_package_json(path: Path, *, allow_version_ranges: bool) -> list[dict]
                     continue
                 version = exact_match.group(1)
 
-            lineno, _ = _npm_dependency_source(txt, section, name)
+            lineno = dependency_lines.get((section, name)) or _find_line(
+                txt, f'"{name}"'
+            )
             deps.append(
                 {
                     "name": name,
--- a/skylos/rules/sca/vulnerability_scanner.py
+++ b/skylos/rules/sca/vulnerability_scanner.py
@@ -536,6 +536,27 @@ def parse_pyproject_toml_candidates(path: Path) -> list[dict]:
     )
 
 
+def _npm_dependency_source(text: str, section: str, name: str) -> tuple[int, str]:
+    in_section = False
+    depth = 0
+    section_key = f'"{section}"'
+    for lineno, raw_line in enumerate(text.splitlines(), 1):
+        line = raw_line.strip()
+        if not in_section:
+            if line.startswith(section_key) and ":" in line:
+                in_section = True
+                depth = line.count("{") - line.count("}")
+                if depth <= 0:
+                    in_section = False
+            continue
+        if line.startswith(f'"{name}"') and ":" in line:
+            return lineno, line
+        depth += line.count("{") - line.count("}")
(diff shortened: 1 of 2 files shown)


mcdigman:
Hi! Thanks for addressing my issue.

I checked this doesn't conflict with my commits and has zero diffs on expanded liveness_primer corpus after my false positive suppression hits; expected because although it *would* have diffs, real popular repositories are unlikely to have good recall anchors for hallucinated dependencies.

Your fix probably works almost as well as it could for a drop-in replacement for _find_line, but it has several edge cases that are hard to fix without a rethink one level further out:
1. Edge case this currently doesn't handle correctly:
```
{
  "keywords": ["lodash"],
  "dependencies": { "lodash": "4.17.20" }
}
```
still anchors on the first line.
2. A more complicated edge case it isn't guaranteed to handle correctly:
```
{
  "dependencies": { "foo": "1.0.0" },
  "overrides": {
    "bar": {
      "dependencies": {
        "foo": "2.0.0"
      }
    }
  }
}
```
With this commit, it will always report `"foo": "2.0.0"`, but in fact that *cannot* be correct given the current skylos implementation, because skylos does not scan overrides currently (that is probably a separate behavioral gap to open an issue about). Technically this case would be a regression from this commit, because it would currently report `"foo":"1.0.0"` purely by happenstance. In practice, it would be better for this function to assume both were scanned and that either or both could be the raising line. 

The deeper issue here is that it is possible for both to say e.g.  `"foo": "1.0.0"`, and for both to be hallucinated, in which case it is simply not possible to determine from the information given which should be matched to which finding.

A real-corpus test case which contains this exact complexity can be found in [pnp/cli-microsoft365/package.json](https://github.com/pnp/cli-microsoft365/blob/3bb75773b655567015f83ad7d79fcfaf7f27451e/package.json#L301), which actually pins the package `uuid` three different times with two different pins in overrides; if those were all hallucinated, it would be basically impossible to match correctly with a function like this.

**Bottom line**
I think the only guaranteed solution to this is to recover the matched line *at the time it is actually matched*. That is annoying, because python standard lib json parser *does not provide any way to do that reliably*. Therefore, it requires either a new dependency, probably `json-source-map` which does that exactly, or hand-rolling a short parser to recover the line numbers immediately *before* the `for section in ("dependencies", "devDependencies"):` in `_parse_package_json`. Because the necessary hand-rolled parser is short, and actual edge cases are probably pretty rare, it may be better not to add the dependency:

Something like (draft function suggested by Opus 5.0, may not be adequately hardened yet):
```
from json.decoder import scanstring

_NPM_DEPENDENCY_SECTIONS = ("dependencies", "devDependencies")
_JSON_STRUCTURAL_RE = re.compile(r'["{}\[\]:,]')


def _npm_dependency_lines(
    text: str, sections: tuple[str, ...]
) -> dict[tuple[str, str], int]:
    """Map (section, dependency name) to the line of its declaring key.

    `text` must already have parsed as JSON; callers run `json.loads` first.

    Walks the manifest once and records a key only while inside a top-level
    object named by `sections`, so the same name appearing anywhere else -- a
    keyword, a description, a nested `dependencies` under `overrides` -- cannot
    hijack the reported line. A later duplicate key overwrites an earlier one,
    matching how `json.loads` resolves duplicates.
    """
    lines: dict[tuple[str, str], int] = {}
    depth = 0
    section: str | None = None
    section_depth = 0
    pending_key: str | None = None
    expect_value = False
    lineno = 1
    scanned = 0
    pos = 0

    while True:
        match = _JSON_STRUCTURAL_RE.search(text, pos)
        if match is None:
            return lines
        char = match.group()
        pos = match.end()

        if char == '"':
            key_start = match.start()
            key, pos = scanstring(text, pos)
            if expect_value:
                pending_key = None
                expect_value = False
                continue
            pending_key = key
            if section is not None and depth == section_depth + 1:
                lineno += text.count("\n", scanned, key_start)
                scanned = key_start
                lines[(section, key)] = lineno
            continue

        if char == ":":
            expect_value = True
            continue
        if char == ",":
            pending_key = None
            expect_value = False
            continue
        if char in "{[":
            if char == "{" and depth == 1 and expect_value and pending_key in sections:
                section = pending_key
                section_depth = depth
            depth += 1
        else:
            depth -= 1
            if section is not None and depth <= section_depth:
                section = None
        pending_key = None
        expect_value = False
```

and then  `lineno = dependency_lines.get((section, name))` where `dependency_lines` is the output of `_npm_dependency_lines`; there are also several other places in this module where `_find_line` could similarly be replaced.

AmirF194:
You're right on both counts, thanks for the detailed walkthrough. Edge case 1 was a real bug: my line-scoped scan closed the section as soon as it saw the open and close brace on the same line, before ever checking that line's content, so it fell through to the first textual occurrence exactly as you described.

Replaced it with a position-based scan close to what you sketched (json.decoder.scanstring over the raw text, tracking depth by absolute position instead of per-line), computed once per file rather than per dependency. Added your through2/keywords and the overrides-nesting case as regression tests; both pass, and the full suite (521 tests touching sca/vulnerability/npm) is unaffected. Pushed as 827a39f, credited you as co-author since the fix follows your sketch closely.

Case 2 comes out fine as a side effect: this scanner never reads `overrides` at all, so the position scan only ever tracks keys directly inside the top-level `dependencies`/`devDependencies` objects and the nested override entry is out of scope by construction, not by luck. Agreed a real overrides scan is a separate issue if it's ever wanted.

mcdigman:
Thanks!

Coverage notes: 

1. The `except` clause at 580 is currently technically unreachable in normal use because json.loads also uses scanstring so if there was going to be a ValueError it is guaranteed to already have been raised in the caller at line 624; I'd be ok keeping that because it is defensive against future drift, and the repo doesn't currently have a 100% line coverage gate. 
2. Similarly, no unit_test reaches the `or _find_line` fallback, and I don't think I can construct one, again because using the identical scanstring makes it difficult for edge cases to diverge; arguably harmless without a 100% branch coverage gate, but worth noting; if other uses of _find_line get similar patches, it could prevent _find_line from being noted as dead.

Other than those non-blocking notes, looks good to me.

**Not caused by PR**

Another preexisting issue I noticed, not induced by your PR:

Elsewhere ([skylos/rules/ai_defect/diff_dependencies.py:75-82](https://github.com/duriantaco/skylos/blob/780432d231e3f92e7b143deff1d194df7da1b786/skylos/rules/ai_defect/diff_dependencies.py#L75)) the repo defines: 
```
_PACKAGE_JSON_DEPENDENCY_SECTIONS = frozenset(
    {
        "dependencies",
        "devDependencies",
        "peerDependencies",
        "optionalDependencies",
    }
)
```

I think skylos should probably be using the exact same section list for consistency here, instead of just `("dependencies", "devDependencies")`; a hallucinated optional dependency can fail silently, so it is *much more likely* to be missed and make it into a merged commit/genuine defect in production code than a hallucinated mandatory dependency. However, that might belong in a follow up issue, along with the scanning overloads issue I mentioned above.

**Edit**
Not technically in the original issue scope, but [_parse_composer_json](https://github.com/duriantaco/skylos/blob/780432d231e3f92e7b143deff1d194df7da1b786/skylos/rules/ai_defect/manifest_dependency_hallucination.py#L565) has the same bug and needs the same fix.

AmirF194:
Thanks for the coverage notes. Agreed on both: the except at 580 and the _find_line fallback are defensive against future drift rather than reachable today, and I would rather keep them than delete coverage-driven dead branches on a repo without a 100% gate.

On optionalDependencies/peerDependencies scope and the _parse_composer_json duplicate: both are real, but they sit in skylos/rules/ai_defect/manifest_dependency_hallucination.py, a different module from this PR's vulnerability_scanner.py, so I would rather send that as its own PR than fold an unrelated file into this one. Will look at it separately; flag me if you get to it first.

duriantaco (approved):
LGTM

AmirF194:
Thanks for the merge, and for working through the edge cases with me.

PR #749, opened 2026-08-21
feat(sca): scan peerDependencies and optionalDependencies in package.json

mcdigman:
Small follow-up commit to #748 based on review/discussion with @AmirF194, adding the two missing dependency sections that are present elsewhere. I didn't open an issue because it is small and already described in the comments, but I can if @duriantaco prefers a strict always-close-a-linked-issue policy.

Note: Arguably, a regular tuple would have been just as good or better here (theoretical but microscopic perf win), but I preserved the frozenset to make it _identical_ to the other frozenset in the codebase.

**AI Use Disclosure**
I designed the change exactly, and Claude Opus 5.0 drafted the unit tests to ensure it landed. A separate Opus 5.0 agent and GPT-5.6 sol reviewed the change according to repository llm review guidelines. I verified with `liveness_primer` on an expanded corpus that the change produces 0 diffs in the current available corpus (expected limitation; well-maintained production codebases should have few recall anchors for hallucinated dependencies).

**Review Note**
Both llm reviewers flagged that currently, this could theoretically flag hallucinated optionalDependencies or peerDependencies declaration that can never _actually_ be installed. My opinion was that flagging in that case is justifiable as more conservative behavior, and that is, if anything, the most plausible way that a hallucinated dependency declaration could subtly creep into a production deployment. 

**Note for @AmirF194**
I had already drafted this when you mentioned interest in doing it; I'm not going to do the override scanning part that I also mentioned right now (its not quite such a micro change) so if that part was still in the scope of what you wanted to do go for it.

## Problem

The SCA scanner only reads `dependencies` and `devDependencies` out of `package.json`. `peerDependencies` and `optionalDependencies` are skipped entirely, so a pinned vulnerable version declared there is never queried against OSV, and the manifest hallucination rule never verifies those names against the npm registry.

This also makes `vulnerability_scanner.py` the odd one out. `skylos/rules/ai_defect/diff_dependencies.py` has walked all four sections since #628, and `skylos/visitors/languages/typescript/workspace.py` walks all four too — only the SCA path was narrow.

## Change

Introduce `_PACKAGE_JSON_DEPENDENCY_SECTIONS`, a frozenset shared by the three places that need the section list:

- `_parse_package_json`, which produces the dependency candidates;
- `_npm_dependency_lines`, the line-anchoring pass added in #748 (it only membership-tests its `sections` argument, so it takes the frozenset directly);
- `_package_json_declaration_count`, which feeds `unresolved_dependency_count` as `declared_count - len(parsed)` and therefore has to count exactly the sections the parser walks.

Those three lists cannot drift apart now. The parse loop iterates `sorted()` so finding order stays stable under hash randomization — `list(frozenset)` of these four names differs from sorted order on every `PYTHONHASHSEED` I tried, so this is load-bearing, and the sorted order preserves the existing `dependencies` → `devDependencies` relative ordering.

## Tests

Four tests over the widened section list, each verified to fail against the pre-change scanner:

- peer and optional entries parse with the right names, versions and line anchors, in the deterministic sorted-section order;
- a name declared in both `peerDependencies` and `optionalDependencies` anchors to its own declaring section rather than the first textual match, including when a section is written on a single line (this is #748's walker doing the work for the new sections);
- `_package_json_declaration_count` counts exactly what the parser walks — the manifest is built from the shared constant, so re-hardcoding a divergent list in either place fails here;
- `unresolved_dependency_count` stays correct end to end when the new sections carry version ranges.

Full suite: 7073 passed, 7 skipped. The two failures in `test_agent_behavior_contract.py` / `test_agent_behavior_openai.py` (`..._reject_excessive_json_nesting_without_crashing`) reproduce identically on unmodified `main` here and are unrelated to this change.

Beyond the suite I ran two 4000-case fuzz passes over randomized four-section manifests (compact, indented, single-line sections, `overrides` decoys, `peerDependenciesMeta`, scoped/escaped/non-ASCII names): every reported line contains its declaring key, repeat parses are identical, and parsed name/version pairs match an independent reimplementation of the section walk and version-resolution rules for both the exact-only and candidate parsers.

## Notes

Worth flagging for review: peer and optional deps are not necessarily installed by the scanned project, and no dep dict in this file carries a section marker, so a peer-dep finding currently reads the same as a runtime one. Also expect `unresolved_dependency_count` to tick up on repos with peer ranges, flipping receipt status to `complete_with_unresolved_versions` more often. Both seem like separate follow-ups rather than something to fold in here — happy to add a `section` field if you'd prefer it in this PR.

Draft because I'd like your read on the peer/optional scanning question above before it's finalized.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skylos/rules/sca/vulnerability_scanner.py
+++ b/skylos/rules/sca/vulnerability_scanner.py
@@ -141,6 +141,16 @@ def _requirements_declaration_count(text: str) -> int:
     )
 
 
+_PACKAGE_JSON_DEPENDENCY_SECTIONS = frozenset(
+    {
+        "dependencies",
+        "devDependencies",
+        "peerDependencies",
+        "optionalDependencies",
+    }
+)
+
+
 def _package_json_declaration_count(text: str) -> int:
     try:
         data = json.loads(text)
@@ -148,7 +158,7 @@ def _package_json_declaration_count(text: str) -> int:
         return 0
     return sum(
         len(section)
-        for key in ("dependencies", "devDependencies")
+        for key in _PACKAGE_JSON_DEPENDENCY_SECTIONS
         if isinstance((section := data.get(key)), dict)
     )
 
@@ -541,7 +551,7 @@ _NPM_JSON_STRUCTURAL_RE = re.compile(r'["{}\[\]:,]')
 
 
 def _npm_dependency_lines(
-    text: str, sections: tuple[str, ...]
+    text: str, sections: frozenset[str]
 ) -> dict[tuple[str, str], int]:
     """Map (section, dependency name) to the 1-indexed line of its declaring key.
 
@@ -625,9 +635,9 @@ def _parse_package_json(path: Path, *, allow_version_ranges: bool) -> list[dict]
     except Exception:
         return deps
 
-    dependency_lines = _npm_dependency_lines(txt, ("dependencies", "devDependencies"))
+    dependency_lines = _npm_dependency_lines(txt, _PACKAGE_JSON_DEPENDENCY_SECTIONS)
 
-    for section in ("dependencies", "devDependencies"):
+    for section in sorted(_PACKAGE_JSON_DEPENDENCY_SECTIONS):
         pkg_deps = data.get(section, {})
         if not isinstance(pkg_deps, dict):
             continue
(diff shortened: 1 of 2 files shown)


mcdigman:
Superseded by #752

PR #750, opened 2026-08-21
fix(dead-code): preserve signature contract parameters

### LINKED ISSUE #723: FP: False positives introduced by PR #658
**Extra Context for this Issue**
This is a follow-up to the example I used as a demo of the value of `liveness_primer` as a CI in #720; there I demonstrated the concept of having an llm analyze the `liveness_primer` flagged corpus diffs introduced between 4.31.0 and 4.31.1. As the llm there flagged, PR #658 actually *did* introduce a huge amount of false positives, which could have been flagged before merge if `liveness_primer` had been integrated as a CI.

**What was flagged?**
- File: Hundreds of sites in almost all the `liveness_primer` corpus sources, pydantic is hit with some of the most notable/egregious false positives, and mitmproxy and pytest have a large absolute number.
- Category: [dead code]

I had Opus 5.0 distill the false positives by #658 and flagged by `liveness_primer` into 6 failure classes cited to actual specific corpus sources, summarized in the attached reproducer. The bulk of them center around, e.g., unused arguments in stub function definitions. 
[skylos_u006_signature_contract_parameters.zip](https://github.com/user-attachments/files/31089131/skylos_u006_signature_contract_parameters.zip)

Here is one of the 6 distilled examples (shape 6), distilled from real pydantic code:

```
class BaseModel:
    """Runtime base supplying the real implementation, as pydantic's does."""

    def model_dump(self, **kwargs: Any) -> Any:
        return dict(kwargs)


class RootModel(BaseModel):
    if typing.TYPE_CHECKING:

        def model_dump(
            self,
            *,
            mode: str = "python",
            include: Any = None,
            exclude: Any = None,
            by_alias: bool | None = None,
        ) -> Any:
            """This method is included just to get a more accurate return type.

            It is included in this `if TYPE_CHECKING:` block since no override
            is actually necessary.
            """
```

gives skylos findings:

```
    █  LOW  dead-code/parameter  Unused parameter: mode
      Dead Code  reproduce.py:186
      mode: str = "python",
      Fix: Remove the unused parameter if it is not public API.

    █  LOW  dead-code/parameter  Unused parameter: include
      Dead Code  reproduce.py:187
      include: Any = None,
      Fix: Remove the unused parameter if it is not public API.

    █  LOW  dead-code/parameter  Unused parameter: exclude
      Dead Code  reproduce.py:188
      exclude: Any = None,
      Fix: Remove the unused parameter if it is not public API.

    █  LOW  dead-code/parameter  Unused parameter: by_alias
      Dead Code  reproduce.py:189
      by_alias: bool | None = None,
      Fix: Remove the unused parameter if it is not public API.
```

The stub functions *themselves* can correctly still be flagged by other skylos if they are unused, or could be flagged by e.g. SKY-L026 incomplete generation rules. *Some* of these could be fixed relatively easily, such as if the target repo's used e.g. explicit PEP 544 Protocols, @abstractmethod decorators, or underscore suppressions that have already been made canonical in skylos; however, requiring these specific patterns may be viewed as noise by users, and these are well-established target repositories; of the patterns identified, only perhaps `shape 5` could be viewed as a true positive relative to how often it occurs in the target repos (e.g., requiring explicit underscore suppression), but it can also be suppressed without a huge extra analysis load, so probably all 6 categories can be handled.

The ideal fix is also *not* fully or partially reverting #658, or leaning more on the grep rescue (except maybe for shape 4). The better follow-up a follow-up to get these false positives in control via the AST. 

Shapes 1., 2., 3, and 6 (can all be solved by the following AST fixes:
1. Extend skip_params at base.py:989 to also skip when the function body is only a docstring/.../pass/raise NotImplementedError
2. Skip params when the method's name matches a method of a resolvable base class — the visitor already tracks abstract_methods (base.py:913) and protocol_method_names, and already exempts explicit @override (base.py:909)

Shape 4 can be mitigated with an AST fix or by allowing the grep recovery to recover that case specifically. 

The shape 5 example distilled from jinja:
```
class RootVisitor:
    def visit(self, node: "Node", **kwargs: Any) -> None:
        getattr(self, f"visit_{type(node).__name__}")(node, **kwargs)

    def visit_AssignBlock(self, node: "Node", **kwargs: Any) -> None:
        for child in node.body:
            self.visit(child)

    def visit_CallBlock(self, node: "Node", **kwargs: Any) -> None:
        for child in node.body:
            self.visit(child, depth=kwargs.get("depth", 0) + 1)


class Node:
    def __init__(self, body: list["Node"] | None = None) -> None:
        self.body = body or []


class AssignBlock(Node):
    pass


class CallBlock(Node):
    pass
```

gives finding
```
    █  LOW  dead-code/parameter  Unused parameter: kwargs
      Dead Code  reproduce.py:144
      def visit_AssignBlock(self, node: "Node", **kwargs: Any) -> None:
      Fix: Remove the unused parameter if it is not public API.
```

and might need its own specific AST fix to handle without an explicit suppression, because it is keeping the interfaces the same for its own walker. We probably don't want to exempt **kwargs specifically in all cases, because **kwargs can be legitimately unused, and that suppression would only address this distilled example specifically, not necessarily all possible similar examples.

If you want, I can draft a fix, and demo the `liveness_primer` analysis of the diffs.

**Skylos version**
skylos 4.33.2

duriantaco:
## Summary

Fixes #723.

PR #658 tightened grep verification for unused parameters, which exposed several signature contract FPs. This adds AST handling instead of loosening grep verification.

This fixes the six cases from the issue:

  - hook and base method stubs
  - abstract method implementations
  - overridden class and factory methods
  - methods declared only for type checking
  - visitor methods that receive forwarded keyword arguments

##Tests

  - All six false positives from the supplied reproducer are gone.

### DIFF
--- a/skylos/analyzer.py
+++ b/skylos/analyzer.py
@@ -58,6 +58,7 @@ from skylos.core.file_discovery import (
 )
 from skylos.core.safe_cache_io import read_project_text_no_symlink
 from skylos.core.linter import LinterVisitor
+from skylos.deadcode.signature_contracts import mark_signature_contract_parameters
 
 from skylos.rules.quality.policy import analyze_repo_policy
 from skylos.rules.vibe_dictionary import build_vibe_dictionary
@@ -4373,6 +4374,7 @@ class Skylos:
             progress_callback(0, 1, Path("PHASE: mark refs"))
         self._mark_refs(progress_callback=progress_callback)
         self._mark_call_arg_method_refs()
+        mark_signature_contract_parameters(self.defs)
 
         if progress_callback:
             progress_callback(0, 1, Path("PHASE: dead-code liveness"))
--- a/skylos/deadcode/signature_contracts.py
+++ b/skylos/deadcode/signature_contracts.py
@@ -1,0 +1,556 @@
+from __future__ import annotations
+
+import ast
+from collections import defaultdict
+from collections.abc import Mapping
+from typing import Any
+
+
+def _shared_signature_parameters(
+    child: ast.FunctionDef | ast.AsyncFunctionDef,
+    parent: ast.FunctionDef | ast.AsyncFunctionDef,
+) -> tuple[set[str], set[str]]:
+    """Return child and parent parameter names occupying shared API slots."""
+    child_names: set[str] = set()
+    parent_names: set[str] = set()
+
+    for child_arg, parent_arg in zip(
+        child.args.posonlyargs, parent.args.posonlyargs
+    ):
+        child_names.add(child_arg.arg)
+        parent_names.add(parent_arg.arg)
+
+    shared_positional_names = {
+        arg.arg for arg in child.args.args
+    } & {arg.arg for arg in parent.args.args}
+    child_names.update(shared_positional_names)
+    parent_names.update(shared_positional_names)
+
+    shared_keyword_names = {
+        arg.arg for arg in child.args.kwonlyargs
+    } & {arg.arg for arg in parent.args.kwonlyargs}
+    child_names.update(shared_keyword_names)
+    parent_names.update(shared_keyword_names)
+
+    if child.args.vararg is not None and parent.args.vararg is not None:
+        child_names.add(child.args.vararg.arg)
+        parent_names.add(parent.args.vararg.arg)
+    if child.args.kwarg is not None and parent.args.kwarg is not None:
+        child_names.add(child.args.kwarg.arg)
+        parent_names.add(parent.args.kwarg.arg)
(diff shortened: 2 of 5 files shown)


mcdigman:
Thanks! Reduces total findings by up to 60% in some repositories. Worth noting that this is also probably a measurable perf win, because less findings earlier means less findings for grep rescue to deal with. Maybe ~10% speedup against the base when merged, though some of that speedup may not stack with my other perf PRs; I didn't fully measure because of that.

PR #767, opened 2026-08-24
perf(analyzer): index abstract override lookups

### LINKED ISSUE #766: Perf: O(n^2) scaling in _check_abstract_overrides

`_check_abstract_overrides` has [several loops which scan over `all_defs`](https://github.com/duriantaco/skylos/blob/46d21dc856342bc0067c6152e45b5576768107cf/skylos/analysis/penalties.py#L948) inside other loops, making them de facto O(n^2) loops; scanning scipy, `_check_abstract_overrides` takes over a minute; on a stress test on another very large public python repo (homeassistant/core/homeassistant), it spends _over half an hour_ in `_check_abstract_overrides` alone, invoking the inner loop elements _billions_ of times.

**Proposed solution**
Pre-compute an index so it is not quadratic

**Alternatives considered**
Some version of building an index is really the only way I can think of, other than leaving it slow.

mcdigman:
Closes #766 

## Summary

- build class and method indexes once for abstract-override checks and cache them

The resulting code is also somewhat more readable IMO

## Measured Performance improvement
Total time in just _check_abstract_overrides:
- scipy/scipy: 65.3 s -> 0.05s
- homeassistant/core/homeassistant: 2058s -> 0.34s

## Validation

- 7,865 passed, 7 skipped, 2 unchanged baseline failures deselected
- 19/19 sandbox-blocked socket tests passed outside the sandbox
- staged Skylos gate passed

**AI Use Disclosure**
Diagnosed with cProfile; codex drafted the initial commit I did the cleanup by hand. Reviewed per repository instructions by Claude Opus 5.0 and GPT-5.6 Sol.

### DIFF
--- a/skylos/analysis/penalties.py
+++ b/skylos/analysis/penalties.py
@@ -881,6 +881,37 @@ def _check_pydantic_methods(def_obj, framework):
     return None
 
 
+def _get_abstract_override_indexes(analyzer, all_defs):
+    cached = getattr(analyzer, "_abstract_override_indexes", None)
+    if cached and cached[0] is all_defs:
+        return cached
+
+    classes_by_file_and_name = {}
+    classes_by_name = {}
+    class_names = set()
+    methods_by_owner_and_name = {}
+    for dobj in all_defs.values():
+        if dobj.type == "class":
+            class_names.add(dobj.simple_name)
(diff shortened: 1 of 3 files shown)


duriantaco (approved):
Mutating analyzer.defs after caching would stale the index, but it looks like the current lifecycle prevents this. just take note. lgtm.

PR #779, opened 2026-08-28
perf(analysis): share parsed ASTs across rule passes

### LINKED ISSUE #775: SKY-L012/SKY-L023: Non-parseable file causes scan loop to silently abort
**Describe the bug**
Currently, if the `ast.parse` in [skylos/analyzer.py:4093](https://github.com/duriantaco/skylos/blob/89d33c0e856f3451eb88c7f60410051ded8d3109/skylos/analyzer.py#L4093) fails (e.g., hits a syntax error), it aborts scanning of the entire `for py_file in _ai_py_files:` loop, keeping only findings for that rule that it had already found before it hit the broken file. Other rules will recover.

Skylos *will* exit nonzero in this case due to the syntax , but there is no indication that this rule specifically reported an arbitrary subset of its findings depending on the loop order. 

The simple fix is to just abort *that particular file*, rather than the entire loop (i.e., best-effort recovery)

**To Reproduce**
Scan `mcp-context-forge` with ai-defects enabled.
`mcp-context-forge/mcp-servers/templates/python/{{cookiecutter.project_slug}}/tests/test_server.py` is a template file that is not parseable by ast.parse as written (maybe could be a separate issue to handle cookiecutter/jinja scaffolding if that is common enough that skylos should have some support for it, I don't know).

It would also happen with any file with an ordinary syntax error, those just aren't very common in well-maintained public repos (unless they are something like a deliberately broken corpus/test file, I guess).

**Expected behavior**
Skip just that file, preserve non-zero exit code

**Actual behavior**
Skips entire rest of rule analysis.

**Environment**
- Python version: 3.14.6

### LINKED ISSUE #778: Perf: duplicated parses and ast walks
There are several duplicated parses and ast walks that could be consolidated. The most obvious candidate (and most performance improving) is that in `api_signature_hallucination.py`, `scan_python_api_signature_hallucinations` parses the entire tree twice, and `_build_parent_map` and `_build_scope_infos` each do two full AST walks, so just in this function two full parses and 4 ast walks could be condensed to just 1 by building an index at the start.

Fixing this roughly halves the time of adding ai-defects, which is the most expensive additional analysis beyond just the dead code findings currently. 

Because fixing that necessarily builds the infrastructure to cache some AST walk results, it can also cache all the AST walk results on the same tree, if desired.

**Proposed solution**
Cache the parses/AST walk results with indices.

**Alternatives considered**
Pre-build the maps and scopes together for just `scan_python_api_signature_hallucinations`, leave it as-is.

mcdigman:
Closes #775 
Closes #778

## Summary

- Shares parsed Python ASTs across repo-wide analysis passes
- reuses a per-tree AST index for phantom-reference and Python API-hallucination analysis, and memoizes scope information shared by both rule passes.
- Fixes a bug where analysis silently aborted on an unparseable file, rather than continuing

## Why?

The ai-defects rules contain 2 full parses and 4 full AST walks that are completely duplicative. 

That could be improved in isolation, and indeed accounts for most of the perf improvement of this PR, but this PR proposes to build indices that unify all the AST walks in the same scope. In measured counts of ast.walk calls, with --ai-defects the PR reduces the total full AST walks from ~12 to ~2. 

I can de-scope unifying all the other AST walks if desired.

As implemented, the fix would inevitably create a merge conflict with an isolated fix to #775, so I just incorporated a fix to that into this PR as well; it could be done as a separate PR that then has its code overwritten by this one, or fixed in isolation if the full unification is not desired.

Also, as implemented, the cache is not thread safe if the outer `analyze` was called separately from multiple different threads. That isn't how skylos uses it so I thought that was an acceptable caveat, but it can be fixed or modified if desired by the maintainer; I didn't want to get *too* far ahead without knowing if merging all the other AST.walks outside ai-defects is ok in the first place. 

**AI Use Disclosure**
The duplicate walks were found by analysis of cProfile results, and scoped the way to unify them. Opus 5.0 drafted the changes and they were reviewed according to skylos repository standards by codex GPT-5.6 Sol and claude Fable 5. 

## Performance impact
About a ~10% reduction in overall time for `liveness_primer` scans including `--ai-defects` and `--quality`. In the mcp-context-forge example, the biggest component of the perf impact is shaving ~30s from the ai-defects scan. For mcp-context-forge, cuts ~31 million calls to ast.walk.

## How to test

- python -m pytest -q test/test_analyzer.py test/test_api_signature_hallucination.py test/test_python_api_hallucination.py test/test_dead_code.py test/test_dead_code_liveness.py
- python -m pytest -q test/
- python scripts/corpus_ci.py --manifest corpus/manifest.json

Results on this head:

- Focused analyzer suite: 275 passed.
- Full suite: 7,900 passed and 7 skipped; two agent JSON-nesting error-message failures reproduce unchanged at base 891f407.
- Corpus guard: 47 cases, 0 failures.
- liveness_primer shows no diffs except the *expected* ones on mcp-context-forge due to the #775 fix.

### DIFF
--- a/skylos/analysis/ast_cache.py
+++ b/skylos/analysis/ast_cache.py
@@ -1,0 +1,212 @@
+"""Run-scoped cache of parsed Python module ASTs.
+
+Repo-wide rule passes (phantom refs, API hallucination coverage, API
+signature checks, dead-code liveness) each need the same files parsed.
+This cache lets them share one tree per file while preserving each call
+site's read semantics: sources are cached per (path, read mode), and a
+tree is parsed once per distinct source text, so call sites with
+different read modes still share a tree whenever the decoded source is
+identical (the common case of a regular, valid-UTF-8 file).
+
+Equal sources are pooled per path, so a file read under several modes
+costs one string rather than one per mode.
+
+Lifetime is a session. Every entry point that populates the cache -- the
+analyzer run and each public rule scanner -- runs inside one, and the
+outermost session clears on the way out, so nothing outlives the call the
+caller made. Sessions nest, so a scanner called from the analyzer still
+shares the run's cache instead of dropping it early. Within a session,
+entries are dropped when a file's identity changes. Derived per-tree
+caches can register a clear callback so their id(tree)-keyed entries
+never outlive the trees stored here.
+
+Not thread-safe: the caches and the session depth are plain module
+globals, so sessions must not overlap across threads -- two entry points
+running concurrently would clear each other's live entries. Skylos'
+own file-level parallelism is process-based, and every caller of these
+entry points is single-threaded.
+"""
+
+from __future__ import annotations
+
+import ast
+import functools
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
+from pathlib import Path
+
+from skylos.core.safe_cache_io import read_text_no_symlink
+
+# Read modes. Each mirrors the exact read semantics of the call sites it
+# replaces; adding a mode must not change an existing one.
+MODE_REPLACE = "replace"  # read_text(encoding="utf-8", errors="replace")
+MODE_STRICT = "strict"  # read_text(encoding="utf-8")
+MODE_SAFE_REPLACE_2MB = "safe_replace_2mb"  # no-symlink read, 2 MiB cap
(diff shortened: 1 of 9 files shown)


duriantaco (approved):
LGTM

PR #780, opened 2026-08-31
fix(analyzer): deterministic module name for multi source-root paths

### LINKED ISSUE #773: Non-deterministic outputs in paths with multiple source root names
**Describe the bug**
When a path contains *multiple* of the qualifiers in
```
_PYTHON_SOURCE_ROOT_NAMES = {"src", "lib", "python"}
```

`_module(self, root, f)`  in `analyzer.py` currently strips one of them, at random, depending on the seed of `PYTHONHASHSEED`. For example, in `IBM/mcp-context-forge`, a finding in 

```
mcp-servers/python/data_analysis_server/src/data_analysis_server/visualization/plots.py
```

Will have its symbol reported as, at random, 

```
data_analysis_server.src.data_analysis_server.visualization.plots.[...]
```

or 

```
 data_analysis_server.visualization.plots.[...]
```

Note that this could also be a problem if, hypothetically, the path had one of the three symbols more than once, like:

```
mcp-servers/src/data_analysis_server/src/data_analysis_server/visualization/plots.py
```

Currently in that case it would only strip the first occurrence.

This behavior was introduced by #361. On affected repos, I think this behavior could potentially induce an *actual* correctness bug down the line, unlike in the sibling issue #733 where the issue was primarily for the benefit of `liveness_primer`'s finding identity.

I think the correct solution in either case is probably to guarantee stripping to the *last* occurrence, e.g. the symbol should be 

```
data_analysis_server.visualization.plots.[...]
```

**To Reproduce**
Run skylos repeatedly on the `mcm-contextforge` file at that path (I think the full exact symbol skylos resolves it to is only reported in the json output?), or use the attached MRE.

[skylos_module_source_root_hash_order.zip](https://github.com/user-attachments/files/31489597/skylos_module_source_root_hash_order.zip)

**Expected behavior**
Finding determinism

**Actual behavior**
Random different finding symbols

**Environment**
- Python version: 3.14.6

DavidPandleton:
## Bug Description

Fixes #773

`_module()` in `analyzer.py` iterates `_PYTHON_SOURCE_ROOT_NAMES` (`{"src", "lib", "python"}`) as a **set**. Set iteration order depends on `PYTHONHASHSEED`, so when a file path contains more than one of these source-root names, the segment that gets stripped is chosen at random per process.

Example from the issue (IBM/mcp-context-forge):

```
mcp-servers/python/data_analysis_server/src/data_analysis_server/visualization/plots.py
```

resolves its symbol randomly to either:

```
data_analysis_server.src.data_analysis_server.visualization.plots...
```

or

```
data_analysis_server.visualization.plots...
```

The same problem occurs when one of the three names appears more than once in a path - only the first occurrence was ever stripped.

## Root Cause

`skylos/analyzer.py` `_module()`:

```python
for source_root_name in _PYTHON_SOURCE_ROOT_NAMES:  # set iteration = hash order
    if source_root_name not in p:
        continue
    source_root_idx = p.index(source_root_name)     # first occurrence only
    ...
```

Non-deterministic output means findings can't be reliably compared across runs (CI vs local, repeated scans), and module-scoped symbol identity is unstable.

## Fix

Scan the path parts right-to-left and strip the **last** non-package source-root directory. This is deterministic regardless of `PYTHONHASHSEED` and matches the expected behavior in the issue (strip to the last occurrence).

## How to Verify

1. `python -m pytest test/test_analyzer.py -k module_name -q` - new regression tests cover both multi-name and repeated-name paths
2. Run `PYTHONHASHSEED=<n> python -c "..."` invoking `_module()` on the repro path with several seeds - result is identical every time

## Test Plan

- [x] Added regression tests: `test_module_name_generation_strips_rightmost_source_root`, `test_module_name_generation_repeated_src_strips_last`
- [x] Existing `test_module_name_generation` still passes
- [x] Full test suite passes

## Risk Assessment

Low - the change only affects how the module name is derived from a path when multiple source-root names are present. Single-source-root layouts (`src/...`, `lib/...`, plain paths) behave identically to before.

### DIFF
--- a/skylos/analyzer.py
+++ b/skylos/analyzer.py
@@ -1260,14 +1260,23 @@ class Skylos:
     def _module(self, root, f):
         p = list(f.relative_to(root).parts)
 
-        for source_root_name in _PYTHON_SOURCE_ROOT_NAMES:
-            if source_root_name not in p:
+        # Strip the rightmost source-root directory (src/lib/python) so the
+        # resulting module name is deterministic regardless of PYTHONHASHSEED.
+        # Scanning right-to-left guarantees a path like
+        # mcp-servers/python/data_analysis_server/src/.../plots.py always
+        # resolves to data_analysis_server... instead of randomly keeping the
+        # first "python" or "src" segment depending on set iteration order.
+        source_root_idx = None
+        for idx in range(len(p) - 1, -1, -1):
+            if p[idx] not in _PYTHON_SOURCE_ROOT_NAMES:
                 continue
-            source_root_idx = p.index(source_root_name)
-            source_root_path = root / "/".join(p[: source_root_idx + 1])
+            source_root_path = root / "/".join(p[: idx + 1])
             if not (source_root_path / "__init__.py").exists():
-                p = p[source_root_idx + 1 :]
+                source_root_idx = idx
                 break
+
+        if source_root_idx is not None:
+            p = p[source_root_idx + 1 :]
 
         for suffix in PYTHON_SIGNATURE_SUFFIXES:
             if p[-1].endswith(suffix):
--- a/test/test_analyzer.py
+++ b/test/test_analyzer.py
@@ -217,6 +217,35 @@ class TestSkylos:
         result = skylos._module(root, file_path)
         assert result == "main"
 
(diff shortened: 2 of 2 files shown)


duriantaco (approved):
LGTM

PR #781, opened 2026-08-31
fix(ai-defect): report all CI permission triggers on one line

### LINKED ISSUE #774: SKY-A103 Randomly drops to only one finding in a line
**Describe the bug**
Sibling of #773 found by an AST walk looking for similar patterns, followed by manual confirmation; that is the only way it likely *would* have been noticed, since by definition this rule could never trigger on a single fixed corpus commit of a repo, only when run as a CI or with local diffs.

[`rules/ai_defect/ci_permission_expansion.py`](https://github.com/mcdigman/skylos/blob/891f407bf5eb7ed2b6a1c453949f0c438f47c803/skylos/rules/ai_defect/ci_permission_expansion.py#L132-L139)

```
def _signal_for_added_line(line: str) -> dict | None:
    for trigger in _PRIVILEGED_TRIGGERS:
        if _line_adds_trigger(line, trigger):
            return {
                "expansion_type": "privileged_trigger",
                "value": trigger,
                "severity": "HIGH",
            }
```

early returns if the line added a trigger, and so is incapable of reporting *multiple* triggers on the same line. Because 
`_PRIVILEGED_TRIGGERS = {"pull_request_target", "workflow_run"}`

is a set, which trigger is chosen is also *random* to due the hash seeding (#773). 

This could be a major issue if a user might plausibly triage adding `pull_request_target` and `workflow_run` at different personal levels of severity, not realizing the other was also present.

**To Reproduce**
Call
```
_signal_for_added_line('on: [pull_request_target, workflow_run]')
```

repeatedly; it will trigger exactly once and at random.

**Expected behavior**
Deterministically trigger on both.

**Actual behavior**
Non-deterministically drops one of the findings.

**Environment**
- Python version: 3.14.6

DavidPandleton:
## Bug Description

Fixes #774

`_signal_for_added_line()` in `rules/ai_defect/ci_permission_expansion.py` early-returns on the first matching privileged trigger, so a single added line can never report more than one finding. Because `_PRIVILEGED_TRIGGERS = {"pull_request_target", "workflow_run"}` is a **set**, *which* trigger gets reported is also random (hash order, same class of bug as #773).

Repro:

```
_signal_for_added_line('on: [pull_request_target, workflow_run]')
```

returns exactly one finding, and the choice is non-deterministic across runs. A user triaging `pull_request_target` at a different severity level than `workflow_run` would never see the second one.

The same "drops to one" behavior exists for inline permissions (`permissions: {contents: write, issues: write}` only reported `contents`).

## Root Cause

`skylos/rules/ai_defect/ci_permission_expansion.py`:

```python
def _signal_for_added_line(line: str) -> dict | None:
    for trigger in _PRIVILEGED_TRIGGERS:   # set iteration = hash order
        if _line_adds_trigger(line, trigger):
            return {...}                    # early return, single finding
```

and `inline_writes[0]` only ever emitted the first inline write permission.

## Fix

Rename to `_signals_for_added_line()` returning a **list** of all signals for the line:

- every matching privileged trigger, iterated in sorted order (deterministic)
- `permissions: write-all`
- every inline `X: write` permission in a `permissions: {...}` block (sorted)
- a standalone `permission: write` line

The caller in `detect_ci_permission_expansion()` loops over the returned signals and emits one finding each (the existing `seen` set still deduplicates across lines).

## How to Verify

1. `python -m pytest test/test_ai_pr_diff_rules.py -q` - new tests cover multi-trigger lines and multi-inline-permission lines
2. Call `_signals_for_added_line('on: [pull_request_target, workflow_run]')` repeatedly (and with different `PYTHONHASHSEED` values) - always returns both triggers, in the same order

## Test Plan

- [x] Added regression tests: `test_detects_all_privileged_triggers_on_same_line`, `test_reports_all_inline_write_permissions`
- [x] Existing tests (`test_detects_github_actions_privileged_trigger_added`, line-move ignore, etc.) still pass
- [x] Full test suite passes

## Risk Assessment

Low - single-trigger and single-permission lines produce the same findings as before. The only behavior change is that multi-trigger / multi-inline lines now report one finding per trigger instead of one random pick.

### DIFF
--- a/skylos/rules/ai_defect/ci_permission_expansion.py
+++ b/skylos/rules/ai_defect/ci_permission_expansion.py
@@ -59,23 +59,20 @@ def detect_ci_permission_expansion(diff_text: str, file_path: str) -> list[dict]
         if not normalized or normalized in removed_normalized:
             continue
 
-        signal = _signal_for_added_line(normalized)
-        if signal is None:
-            continue
-
-        key = (signal["expansion_type"], signal["value"])
-        if key in seen:
-            continue
-        seen.add(key)
-        findings.append(
-            _make_finding(
-                file_path,
-                line.line_no,
-                expansion_type=signal["expansion_type"],
-                value=signal["value"],
-                severity=signal["severity"],
+        for signal in _signals_for_added_line(normalized):
+            key = (signal["expansion_type"], signal["value"])
+            if key in seen:
+                continue
+            seen.add(key)
+            findings.append(
+                _make_finding(
+                    file_path,
+                    line.line_no,
+                    expansion_type=signal["expansion_type"],
+                    value=signal["value"],
+                    severity=signal["severity"],
+                )
             )
-        )
 
     return findings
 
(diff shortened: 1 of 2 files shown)


duriantaco:
Closing as superseded by #792 & #793. Fix is already in main, with more edge cases coverage. Thanks for your contribution! Appreciate it !

PR #782, opened 2026-08-31
fix(catalog): swap SKY-F101/F102 rule descriptions

DavidPandleton:
catalog.py listed F101 as 'Framework route missing auth' and F102 as 'Framework handler practice issue', but practices.py emits F101 for missing response contracts and F102 for missing auth guards on mutating routes. Swap the names to match actual behavior.


duriantaco (approved):
Reviewed and approved. thanks!

PR #783, opened 2026-08-31
fix(jwt): detect decode calls through import aliases and direct imports

DavidPandleton:
_JWTChecker only matched  and .  and  both bypassed detection. Track import aliases and direct function imports from the jwt module, then check against them in visit_Call.


duriantaco (approved):
sflr. added scoping cleanup and reachability handling. other than that, lgtm. thanks!

PR #784, opened 2026-08-31
fix(cwe): correct 10 misaligned CWE mappings

DavidPandleton:
10 CWE mappings in standards.py pointed at unrelated CWEs. L001 was mapped to CWE-1321 (JS prototype pollution) for a Python mutable-default-args rule; L027 to CWE-1095 (loop condition) for duplicate string literals; L028 to CWE-1075 (finally-block) for too-many-returns; Q401 to CWE-667 (improper locking) for async blocking. Corrected those and 6 others to CWEs that match each rule's actual behavior.

### DIFF
--- a/skylos/rules/quality/standards.py
+++ b/skylos/rules/quality/standards.py
@@ -4,8 +4,8 @@ CWE_MAP: dict[str, list[dict[str, str]]] = {
     # Logic rules
     "SKY-L001": [
         {
-            "id": "CWE-1321",
-            "name": "Improperly Controlled Modification of Object Prototype Attributes",
+            "id": "CWE-665",
+            "name": "Improper Initialization",
         }
     ],
     "SKY-L002": [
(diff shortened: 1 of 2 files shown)


duriantaco (approved):
Reviewed the final two-file diff at 0df361a. The corrected CWE mappings match detector behavior, focused validation passed, and all fresh CI checks are green.

duriantaco (commented):
SKY-L027: use CWE-1106.. the rule detects repeated literals, not duplicated implementations.

SKY-L028: use CWE-1119; the rule counts return statements, not McCabe complexity.
SKY-Q401: use CWE-710 .. detector does not prove uncontrolled resource consumption.
CWE-1064 should remain mapped to SKY-C303, which detects excessive parameters. 

Please also update the regression assertions to check both CWE IDs and official names.

Approved

PR #805, opened 2026-09-06
fix(security): avoid SKY-D212 RegExp literal false positives

### LINKED ISSUE #803: FP: SKY-D212 on RegExp.exec() — receiver check is a variable-name allowlist
**What was flagged?**

- File: `app.ts`
- Name: `codeBlockRegex.exec(rawContent)`
- Category: security

```
app.ts:6  SKY-D212  child_process.exec() can lead to command injection. Use execFile() instead.
```

`codeBlockRegex` is a regex literal. `.exec()` is `RegExp.prototype.exec`. There is no `child_process` import in the file.

The guard is a name allowlist, `_SAFE_EXEC_OBJECTS` at `skylos/visitors/languages/typescript/danger.py:34`: `regex`, `re`, `regexp`, `pattern`, `reg`, `db`, `stmt`, `query`, `statement`, `cursor`, `conn`, `connection`. Anything outside those twelve names gets flagged. The rule never looks at what the variable holds.

**Skylos version**

4.35.0

**Minimal reproduction**

`app.ts`:

```ts
const codeBlockRegex = /x(y)?z/gi;

export function findBlocks(rawContent: string): string[] {
    const blocks: string[] = [];
    let match: RegExpExecArray | null;
    while ((match = codeBlockRegex.exec(rawContent)) !== null) {
        blocks.push(match[1]);
    }
    return blocks;
}
```

```bash
skylos . --danger --format concise --no-provenance --no-upload
```

```
app.ts:6  SKY-D212  child_process.exec() can lead to command injection. Use execFile() instead.
```

Rename it `regex` and the finding disappears. Rename it `matcher` and it stays.
Checked via `scan_danger` on 4.35.0:

| receiver | result |
|---|---|
| `regex` | clean |
| `pattern` | clean |
| `codeBlockRegex` | SKY-D212 |
| `matcher` | SKY-D212 |

Fixing it properly means checking the binding's initializer (regex literal or `new RegExp(...)`), not the name. `_child_process_aliases` in the same file already works that way — off actual `require("child_process")` evidence.

mcdigman:
## What does this PR do?

Closes #803 

Hi @duriantaco, I see you already self-assigned the new issue #803; however, I happen to already have a fix built for it, because this is one of the two remaining false positives when scanning skylos on `liveness_primer` itself. I think I made it before I had made any other PRs and I guess I forgot to PR it or actually open an issue before; feel free to close this if you don't want this solution/already wrote a better one.

## Summary
- Determine if a javascript `.exec` call is actually a regular expression regardless of the name of the variable

## Why?

SKY-D212 currently reports ordinary RegExp calls when the binding has an application-specific name:

`explorer/src/lib/permalink.js:34 SKY-D212 child_process.exec() can lead to command injection`

The old suppression was based on names such as regex and pattern, so it both missed this safe GITHUB_PATTERN.exec(url) call and could be evaded by assigning a dangerous receiver to a safe-looking name.

With this change, the same file produces no SKY-D212 finding while unsafe lookalikes remain reported.

## Precision Impact

I think it strictly improves precision.

## Checklist

- [x ] Tests pass (`python3 -m pytest test/`)
- [x ] No new false positives introduced (if modifying analysis logic)
- [x] Added or updated a corpus case for any confirmed precision regression or false positive fix
- [x ] Ran the corpus guard (`python3 scripts/corpus_ci.py --manifest corpus/manifest.json`) if analysis logic changed

### DIFF
--- a/skylos/visitors/languages/typescript/danger.py
+++ b/skylos/visitors/languages/typescript/danger.py
@@ -611,12 +611,20 @@ def _child_process_aliases(root_node, lang: Language, source_bytes: bytes) -> se
     return aliases
 
 
+def _field_matches_node(
+    node: Node | None, field_name: str, expected_node: Node
+) -> bool:
+    if node is None:
+        return False
+    field_node = node.child_by_field_name(field_name)
+    return field_node is not None and field_node.id == expected_node.id
+
+
 def _regexp_use_kind(node: Node, source_bytes: bytes) -> str | None:
     member_node = node.parent
     if member_node is None or member_node.type != "member_expression":
         return None
-    object_node = member_node.child_by_field_name("object")
-    if object_node is None or object_node.id != node.id:
+    if not _field_matches_node(member_node, "object", node):
         return None
 
     property_node = member_node.child_by_field_name("property")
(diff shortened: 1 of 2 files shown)


duriantaco:
Hey thanks @mcdigman #806 includes your original commits and regression tests, made some changes on top of yours. Closing this as superseded by #806. Appreciate your help on this 👍

PR #834, opened 2026-09-11
fix(circular): skip fallback self-cycles

### LINKED ISSUE #833: FP: SKY-CIRC on generated files
**What was flagged?**
- File: `cosmic._commit_hash`, `platformdirs.version`, `pandas._libs`, `pandas._version_meson`. 
- Category: [circular import]

The above files are all generated files that either do not exist at analysis time or are in `.gitignore`. In those cases, skylos currently removes the submodule name, and then reports it as a circular import; e.g. `pandas -> pandas` or `cosmic -> cosmic`. 

The circular import finding in this case is:

1. Not actually a circular import
2. Any diagnostic information is deliberately stripped, so the alert message is just an unhelpful `cosmic -> cosmic`, rather than any actionable information for an end user

There is currently a unit test that actually asserts this behavior for unresolved submodules, but I think that is wrong. Some of these cases could potentially be covered by some *different* analysis rule about imports, but it is unclear to report the core issue as being a circular dependency.

Alternatively the `version`-like files could be resolved via a more sophisticated build-system-aware rule that was aware of files that `Hatch`, `meson`, or `py.extension_module()` might generate, but that would be a more brittle and larger change because it would have to be aware of several different kinds of build systems.

**Skylos version**
skylos 4.36.1

**Minimal reproduction**
Package structure distilled from the above cases with false positives and control cases attached. 

[skylos_circ_unresolved_child_fallback.zip](https://github.com/user-attachments/files/32124970/skylos_circ_unresolved_child_fallback.zip)

mcdigman:
Closes #833 

## Summary

- suppress a circular-graph self-edge only when a more-specific unresolved child falls back to the importing module
- keep the conservative fallback in the architecture graph, while preserving exact self-imports and unresolved base-module symbols
- cover absolute and relative imports in both AST and raw-import analysis
- Revise unit test to cover new proposed behavior

## Validation

- `pytest -q test/test_circular_deps.py` — 72 passed, 2 skipped
- `pytest -q test/test_architecture.py` — 35 passed
- full suite — 9,856 passed
- Ruff on the touched files — unchanged at the same 60 pre-existing findings on fork `main`
- pinned reproductions:
  - COSMIC `_commit_hash`: 1 self-cycle before, 0 after
  - platformdirs generated `version`: 1 self-cycle before, 0 after
  - pandas compiled/generated children at `_libs/__init__.py:16` and `pandas/__init__.py:178`: 2 self-cycles before, 0 after

## AI Use Disclosure
Fix to discovered false positives conceived with help of Codex GPT-5.6 Sol to cover the identified false positive cases in a defensible manner as narrowly as possible. Reviewed according to skylos repository llm guidelines by a separate GPT-5.6 Sol agent and Opus 5.0. I inspected the changes myself.

### DIFF
--- a/skylos/analysis/circular_deps.py
+++ b/skylos/analysis/circular_deps.py
@@ -84,7 +84,7 @@ def _circular_import_targets(
         return {
             target: target_names
             for target, target_names in targets.items()
-            if target != from_module or import_module == target
+            if target != from_module or not import_module.startswith(f"{target}.")
         }
     return {root: names}
 
--- a/skylos/analysis/circular_deps.py
+++ b/skylos/analysis/circular_deps.py
@@ -74,13 +74,18 @@ def _circular_import_targets(
     Collapsing ``pkg -> pkg.child`` to ``pkg -> pkg`` invents a self-cycle,
     while discarding that edge would hide a real child-to-package cycle.
     Keep the exact resolved graph within a package; unresolved symbols still
-    fall back to their known containing module.
+    fall back to their known containing module. Do not turn a more-specific
+    unresolved child that falls back to its importer into a self-cycle.
     """
     if not targets:
         return {}
     root = _module_root(import_module)
     if _module_root(from_module) == root:
-        return targets
+        return {
+            target: target_names
+            for target, target_names in targets.items()
+            if target != from_module or import_module == target
+        }
     return {root: names}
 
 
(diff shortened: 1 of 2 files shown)


duriantaco (approved):
Approved. Reproduced #833 and verified that the FPs are gone.

Also checked that real circular imports are still detected, and the fake imports no longer affect results.

PR #838, opened 2026-09-11
fix(circular): make cycle selection independent of iteration order

### LINKED ISSUE #837: SKY-CIRC: Nondeterministic findings
**Describe the bug**
SKY-CIRC does a pruned DFS to identify import cycles. It is necessary to not report every single possible circular finding, because in repos like `pandas` which genuinely have a large number of overlapping circular loops, the overlapping loops create exponentially increasing amounts of reported cycles, which it is not very useful to report exhaustively. 

However, if the DFS is in a non-deterministic order, the non-exhaustive search means that there can be huge churn in the number of findings; for `pandas`, typical run-to-run variance in the SKY-CIRC findings is O(1,400) different findings (including both added and dropped findings as changes). That variance would make processing skylos results to fix the findings deterministically difficult. It also blocks adding `SKY-CIRC` to `liveness_primer` finding identifies. For example, using `liveness_primer` to run skylos on pandas twice with *identical commit SHAs*:

```
summary
base findings 1152, head findings 1133
totals: 698 new, 717 dropped, 0 changed (0 confidence-only, 0 message-only, 0 severity-only)
new 698: SKY-CIRC 698
dropped 717: SKY-CIRC 717
```

Skylos *itself* also has a non-deterministic set of circular imports findings. 

The fix is to pick a canonical DFS search order.

**To Reproduce**
Example package (`fixtures`) with non-deterministic repeated scans attached. 

[skylos_circular_dependency_hash_seed_nondeterminism.zip](https://github.com/user-attachments/files/32131115/skylos_circular_dependency_hash_seed_nondeterminism.zip)

**Expected behavior**
Deterministic results

**Actual behavior**
Nondeterministic results

**Environment**
- OS: [e.g. macOS 14, Ubuntu 22.04]
- Python version: [e.g. 3.11]
- Skylos version: [e.g. 4.1.1] (`skylos --version`)

**Additional context**
Any other context — screenshots, logs, related issues.

mcdigman:
Closes #837

## Summary

-  Iterate neighbors and roots in sorted order:

   - `_find_cycles_py`: `sorted(...)` on neighbors and roots. 
   - `_find_cycles_fast`: hand the native finder sorted edges and modules.
   -  `rust/src/cycles.rs`: mirror changes so parity tests in `test/test_fast_parity.py` pass.

- sort findings by (len(cycle), cycle) so display order is canonical too.
- I verified `liveness_primer` results are deterministic with this fix, so it will un-block adding SKY-CIRC to `liveness_primer`

Not changed: this does not enumerate all elementary cycles (exponential on dense import SCCs), and it keeps the existing dedup by node set.

Should not have a merge conflict with #834 despite touching the same files, and so could be merged in either order.

## Verification

- Reproduction script before: 4 or 5 cycles depending on seed. After: 24 unseeded runs and seeds 0 through 7 all report the same 5 cycles.
- `ruff check` finding count and `ruff format --check` unchanged from `main` for the touched files.

## AI Use Disclosure
Fixes drafted by Claude Fable 5.1, and reviewed per skylos llm repository review guidelines by Codex GPT-5.6 Sol and Claude Opus 5.0. I reviewed the changes myself, and verified they create the expected deterministic output.

### DIFF
--- a/rust/src/cycles.rs
+++ b/rust/src/cycles.rs
@@ -22,6 +22,14 @@ pub fn find_cycles(
     for (from, to) in &edges {
         deps.entry(from.clone()).or_default().push(to.clone());
     }
+    // Sorted traversal: the result is a function of the graph, not of the
+    // order in which the caller happened to enumerate edges and modules.
+    for neighbors in deps.values_mut() {
+        neighbors.sort();
+        neighbors.dedup();
+    }
+    let mut modules = modules;
+    modules.sort();
 
     let mut all_cycles: Vec<Vec<String>> = Vec::new();
 
--- a/rust/src/cycles.rs
+++ b/rust/src/cycles.rs
@@ -4,7 +4,11 @@
 use pyo3::prelude::*;
 use std::collections::{HashMap, HashSet};
 
-/// Find all circular dependency cycles using DFS (matches Python's find_simple_cycles exactly).
+/// Find circular dependency cycles using DFS (matches Python's find_simple_cycles exactly).
+///
+/// Like the Python finder this reports one cycle per back edge of a pruned
+/// traversal, not every elementary cycle; the sorted traversal only makes
+/// that selection a deterministic function of the graph.
 ///
 /// Args:
 ///     edges: List of (from_module, to_module) import edges.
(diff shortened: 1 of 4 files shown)


duriantaco (approved):
Results now consistent across diff hash seeds. Both python and rust implementations match. LGTM. Thank you!

PR #853, opened 2026-09-15
fix(typescript): fold static esbuild entry points

### LINKED ISSUE #852: FP: More esbuild entry point SKY-E003 cases
**What was flagged?**
- File: `explorer/src/app/worker.js`
- Name: `liveness_primer`
- Category: [dead code]

One of the two esbuild entry point false positives from #836 still affects `liveness_primer` after #846. The cases is like (distilled example):

```
const here = dirname(fileURLToPath(import.meta.url));
const shared = { bundle: true };

await build({
  ...shared,
  entryPoints: [join(here, 'src', 'app', 'worker.js')],
});
``` 

which is flagged because skylos cannot recognize the static join. This exact case could be fixed directly relatively easily, though I scanned through some other repos that use esbuild and found some other similar false positive cases:

From `github/codeql-variant-analysis-action`: 
```actions.map(n => \src/${n}.ts)```

from `Brawl345/CNL-decryptor` and others:
```path.resolve(__dirname,'source','service-worker.ts')```

and others, included in the attached reproducer package structure. 

Most of the examples (6 of the 8) in the reproducer can be fixed with essentially the same tree sitter fix; skylos could conservatively abstain from reporting in the other 2 less common false positive cases without resolving them exactly, or just allow them for now because they are harder to handle. 

**Skylos version**
skylos 4.37.0

**Minimal reproduction**

[skylos_e003_dynamic_entrypoints.zip](https://github.com/user-attachments/files/32262628/skylos_e003_dynamic_entrypoints.zip)

mcdigman:
Closes #852 

## Summary

- fold module-level const strings, templates, arrays, and object spreads in esbuild entry points
- resolve proven `node:path` and `node:url` helpers plus literal-array `map` callbacks
- refactor tree-sitter helpers into shared `js_ast.py`

Doesn't handle the two cases found in a claude scan of esbuild repos that can't be handled by this simple fix; the easiest way to handle those would probably be to just abstain from reporting E003 findings that *could* match without verifying whether they do for certain, or to just ignore them unless they become a downstream problem.

## Validation

- focused TypeScript suite: 383 passed, 4 skipped
- Verified the lingering `liveness_primer` false positive is handled
- Verified the 6 expected fixed cases in the reproducer hit

## AI Use Disclosure
Fix drafted by Claude Opus 5.0 and reviewed according to skylos repository instructions by Astra 6.0 and a separate Opus 5.0 agent. I reviewed the code; the `liveness_primer` case specifically could easily be handled with a simpler fix, but this fix reduces more false positive cases.


duriantaco (approved):
LGTM

PR #858, opened 2026-09-16
fix(grep-verify): keep rescue evidence inside the analyzer scan boundary

### LINKED ISSUE #855: Grep verification ignores --exclude / .gitignore, so a large ignored dir exhausts the budget and all dead-code findings are withheld
**Describe the bug**

Grep verification does not honor the analyzer's exclusions. The analyzer skips `--exclude` folders, its default excludes, and (since #125) `.gitignore`d paths, but the grep rescue pass walks the git root with its own fixed list (`_GREP_EXCLUDE_DIRS` in `skylos/core/grep_verify_common.py`) and, on the ripgrep backend, `--hidden --no-ignore`. Nothing from `exclude_folders` reaches `grep_verify_findings()` (call site in `analyzer._grep_verify`, still true on `main` @ 480faa4).

On a repo with a large ignored directory that is not in that fixed list, every candidate's grep is run over the ignored tree too. In my case the ignored directory is `.claude/worktrees/` (Claude Code worktrees: full copies of the project), and it makes the pass exhaust a 300 s budget on a 574-file Python project, so dead-code results are withheld entirely (`SKY-ANALYSIS-INCOMPLETE`, `affected_file_count: 247`, 537 candidates abstained). Passing `--exclude .claude` has no effect on the grep pass.

**To Reproduce**

```bash
pip install skylos==4.37.0
cd <django-repo>            # 574 .py files in the project proper
mkdir -p .claude/worktrees && echo '.claude/' >> .gitignore
# populate .claude/worktrees with ~15 checkouts of the same repo (8,490 .py files in total)
skylos . --exclude .claude --format concise
SKYLOS_GREP_BUDGET=300 skylos . --exclude .claude --format concise
```

**Expected behavior**

Excluded and ignored folders are excluded from grep verification as well, so the pass sees the same 574 files the analyzer analysed and completes within the default budget.

**Actual behavior**

Both runs end with only `SKY-E002`/`SKY-E003` findings plus:

```
ai/providers/base.py:82  SKY-ANALYSIS-INCOMPLETE  Grep verification exceeded its 300-second budget. Dead-code findings that required grep verification were withheld. Increase SKYLOS_GREP_BUDGET and rerun.
```

`--json` → `analysis_summary.dead_code_evidence.candidate_decisions = {"reported": 0, "rescued": 20, "abstained": 537}`.

Measured per-pattern cost of the exact command skylos builds (`_direct_grep_command`, system-grep backend since `rg` is not on PATH here):

| command | files with a hit | wall time |
|---|---|---|
| as built by skylos (fixed exclude list) | 291 | 1.40 s |
| same + `--exclude-dir .claude` | 18 | 0.07 s |

537 candidates × 1.4 s ≈ 750 s, hence the timeout; with the exclusion honored it would be ≈ 40 s. The ripgrep backend has the same gap by construction (`--hidden --no-ignore` + the same fixed list), it is just batched so the wall-clock impact is smaller. 291 vs 18 files also means the evidence itself is polluted: usage in a stale worktree copy would "rescue" a symbol that is dead in the real tree.

**Environment**

- OS: macOS 14.3 (Darwin 23.3.0), arm64
- Python version: 3.13.14
- Skylos version: 4.37.0
- `rg` not on PATH (`shutil.which("rg") is None`), so `_trusted_which` selects `/usr/bin/grep`

**Additional context**

- Related: #125 (analyzer honors `.gitignore`, grep pass does not), #714 / #722 (budget behaviour and batching; this is the case where the tree itself is 15× larger than it should be).
- Suggested fix: pass the analyzer's resolved `exclude_folders` (and the gitignore-derived set) into `grep_verify_findings()` and append them as `-g '!**/<dir>/**'` / `--exclude-dir` alongside `_GREP_EXCLUDE_DIRS`; or drop `--no-ignore` on the ripgrep backend.
- Workaround for others hitting this: none that keeps grep verification; `--no-grep-verify` finishes in 18 s but reports string-referenced Django symbols (context processors, model properties) as dead, so it is not a substitute.

mcdigman:
Closes #855 

## Summary

- don't scan ignored or excluded directories with grep verification
- filter grep rescues from repos that were ignored or excluded
- add `rg` to the CI workflow so the new tests can actually run

## Verification

- fixes all reproducer cases I posted in the comments
- performance is improved with explicit excludes

Performance on `mitmproxy` with three full copies of itself placed under a gitignored `.claude/worktrees/`:

| variant | grep-verify phase |
| --- | --- |
| main | 6.78 s, 6.94 s |
| this branch | 2.72 s, 2.69 s |

## AI Use Disclosure
Fix drafted by Claude Fable 5.1, reviewed according to skylos repository llm review guidelines by GPT-5.6 Sol and a separate Claude 5.1 agent. I reviewed the code.


duriantaco (approved):
Thanks @mcdigman for working on this! I pushed a small follow-up in 7728968, which is a problem where an ignored folder sharing the checkout’s name could make grep skip the entire project. The rest LGTM.

PR #869, opened 2026-09-18
feat(cli): add --format json-ci for compact CI/agent JSON output

### LINKED ISSUE #866: perf(cli): --format json emits a 46MB payload where 63.6% is dead-code evidence for symbols that are not findings
**Describe the bug**

`--format json` writes a top-level `dead_code_evidence` block that enumerates every analysed symbol, including those classified `alive`. On a mid-size Python package (`./skylos`, commit `82d463a`): 42,936 evidence entries for 96 reported findings (447:1); 99.0% of the entries are `alive`. The output file is 46.4 MB, of which `dead_code_evidence` is 30.98 MB (63.6%).

`docs/cli-output.md` presents `--format json --output agent-results.json` as the machine-consumable path for agents and CI, and a consumer parsing this file peaks at ~202 MB RSS (4.3x the file size).

**To Reproduce**

```bash
pip install skylos==4.38.0
git clone --depth 1 https://github.com/duriantaco/skylos && cd skylos
skylos ./skylos --no-provenance --format json --output out.json
stat -c %s out.json          # 48,704,846
```

Measure the split:

```python
import json
d = json.load(open("out.json"))
ev = d["dead_code_evidence"]["symbols"]
print(len(ev))   # 42936
# by classification: alive=42517, uncertain=245, likely_dead=174
```

**Expected behavior**

A machine-readable report scales with the findings rather than the full symbol table. Symbols classified `alive` are not findings.

**Actual behavior**

- 42,936 evidence entries for 96 findings (**447:1**); 99.0% classified `alive`.
- No flag omits the block — checked `--help` for slim/compact/minimal/no-evidence/summary-only/quiet; none exists.
- The `summary.pop("dead_code_evidence", None)` calls at `cli.py:1759` and `cli.py:1842` strip it from `analysis_summary` on the `--select`/category-filter paths only; the default path emits the full top-level block.

**Environment**

- OS: Ubuntu (Linux 6.17.0-1022-azure)
- Python: 3.12.14
- Skylos: 4.38.0 (commit `82d463a`)

**Additional context**

A question before any fix: besides the TUI (`skylos/ui/dead_code_evidence.py:121`), are there other consumers that rely on the `alive` entries? If so, restricting the block would break them.

Options, offered not prescribed:

1. Restrict `dead_code_evidence` to non-`alive` classifications (174 + 245 here) — smallest change.
2. Gate the full ledger behind `--verbose`, emit counts by default.
3. Add an explicit opt-out flag.

If the full ledger in the default payload is intentional design, options 1/2 are moot — but option 3 would still let CI/agent consumers opt out of the cost. Happy to implement whichever direction you prefer.

blrain3:
## What does this PR do?

Adds a new `--format json-ci` output mode that emits findings without the two
bulk keys that make `--format json` unusable in CI and agent pipelines.

On a scan of this repo the default JSON is dominated by two keys:

| key | size |
| --- | --- |
| `dead_code_evidence` | ~31 MB |
| `definitions` | ~17 MB |

`json-ci` omits both, at top level and wherever they appear nested (including
`analysis_summary.dead_code_evidence`). Everything else is kept:

- `findings` (all finding kinds)
- per-finding evidence (each finding keeps its own `dead_code_evidence` field)
- summary counts (`total_files`, `total_loc`, per-kind counts, `analysis_error_count`)
- `analysis_errors`
- exit behavior (unchanged)

The default `--format json` output path is untouched. No new opt-out flag is
added — compact output is opt-in only.

Note: this PR does not modify the analysis result itself. Only the serialized
display copy is trimmed; cloud upload / TUI paths still receive the full
result object.

## Why?

Closes #866, following the direction agreed in the issue thread.

## How to test

```bash
skylos . --format json     > /tmp/default.json
skylos . --format json-ci  > /tmp/ci.json
```

On this repo: `/tmp/default.json` is **48,706,315 bytes**, `/tmp/ci.json` is
**290,057 bytes** — about 1/168 the size (99.4% smaller).

Confirm `json-ci` no longer has the bulk keys while keeping findings:

```bash
python3 -c "
import json
d=json.load(open('/tmp/ci.json'))
print('top keys:', sorted(d))
print('findings:', len(d['findings']))
print('all have evidence:', all('dead_code_evidence' in f for f in d['findings']))
"
```

## Default output unchanged

The default `--format json` path is unchanged at the logic level: structure and
values match the pre-change baseline.

One caveat worth stating up front, because it is easy to mistake for a
regression: this project's JSON output is **not byte-stable across runs** — it
has non-deterministic set iteration order. Running the same
`--format json` twice on an unchanged tree produces differing key order in
`analysis_summary.excluded_folders` and `calls`. I verified this on unmodified
upstream code: 47 of 61 entries differed between two consecutive runs with this
PR's changes stashed. So the baseline comparison normalizes set-derived
ordering rather than comparing raw bytes; anyone diffing raw output will see
noise that predates this PR.

## Test evidence

```
12731 passed, 2 failed, 48 skipped
```

The 2 failures are pre-existing and unrelated. Verified by stashing this PR's
changes and running the same two tests on unmodified upstream code — both fail
identically:

```
FAILED test/test_dead_code_benchmark.py::test_checked_in_dead_code_benchmark_passes
FAILED test/test_security_benchmark.py::test_checked_in_security_benchmark_passes
```

New focused tests: `test/test_cli_json_ci_format.py` (14 tests, all passing),
covering the omit list, preserved findings/evidence/counts, default-output
invariance, and SARIF (`--sarif`) still working.

## Precision Impact

None. No analysis logic, rule, or detection behavior is changed — this is
output serialization only. No corpus case is affected.

## Checklist

- [x] Tests pass (`python3 -m pytest test/`) — 12731 passed, 2 pre-existing failures
- [x] No new false positives introduced (no analysis logic changed)
- [x] No corpus case needed (no precision change)
- [x] No corpus guard run needed (no analysis logic changed)
- [x] CHANGELOG — generated by release tooling from commit messages; not hand-edited

## AI Use Disclosure

Drafted with DeepSeek assistance (deepseek-v4.1-flash), following the direction
agreed in #866. I specified the scope, reviewed every line, ran the full
verification, and can explain each part. The measurements above are from my own
runs, not generated estimates.


duriantaco (approved):
Thanks, one issue in the PR was that it also removed `analysis_summary.dead_code_evidence`. That field actually contains small summary counts, not the large symbol list. So we gotta keep it. 
 
I pushed a follow-up commit that removes only the two large top level fields and restored the summary counts.  Otherwise LGTM. Thanks for contributing! Appreciate it 👍

